PR #12, opened 2024-01-03
Utilities for accessing validator history state

ebatsell:
Allows easy access of fields without needing to do range math or default checks when computing on validator history state

### DIFF
--- a/programs/validator-history/src/state.rs
+++ b/programs/validator-history/src/state.rs
@@ -134,9 +134,7 @@ macro_rules! field_latest {
 
 macro_rules! field_range {
     ($self:expr, $start_epoch:expr, $end_epoch:expr, $field:ident, $type:ty) => {{
-        let epoch_range = $self
-            .epoch_range($start_epoch, $end_epoch)
-            .unwrap_or_default();
+        let epoch_range = $self.epoch_range($start_epoch, $end_epoch);
         epoch_range
             .iter()
             .map(|entry| {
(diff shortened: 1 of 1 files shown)


programs/validator-history/src/state.rs
```diff
@@ -148,6 +180,69 @@ impl CircBuf {
     pub fn arr_mut(&mut self) -> &mut [ValidatorHistoryEntry] {
         &mut self.arr
     }
+
+    pub fn epoch_range(
+        &self,
+        start_epoch: u16,
+        end_epoch: u16,
+    ) -> Option<Vec<&ValidatorHistoryEntry>> {
```
buffalu, line 188: another option: Vec<Option<&ValidatorHistoryEntry>>. 

what if epoch missing? for averaging, i ask for 5, get 2. i assume you use x.epoch_range(...).len() for avg?

programs/validator-history/src/state.rs
```diff
@@ -148,6 +180,69 @@ impl CircBuf {
...
+        // Returns &ValidatorHistoryEntry for each existing entry in range [start_epoch, end_epoch], factoring for wraparound
```
buffalu, line 189: can you add this as docstring with /// above fn?

buffalu: also add note about up to AND including end_epoch

programs/validator-history/src/state.rs
```diff
@@ -454,4 +549,53 @@ mod tests {
...
+        // same start and end epoch
+        let epoch_range = circ_buf.epoch_range(0, 0).unwrap();
+        assert_eq!(epoch_range.len(), 1);
+
+        // Test start epoch out of range
+        let epoch_range = circ_buf.epoch_range(4, 3);
+        assert!(epoch_range.is_none());
+
+        // Test end epoch out of range
+        let epoch_range = circ_buf.epoch_range(0, 5);
```
buffalu, line 594: imo this should prolly return [0,1,2,3,4]? what if you don't have the most recent epoch?

programs/validator-history/src/state.rs
```diff
@@ -454,4 +549,53 @@ mod tests {
...
+        assert!(epoch_range.is_none());
+
+        // None if start epoch is none
+        let epoch_range = circ_buf.epoch_range(2, 3);
```
buffalu, line 598: same here, seems like it should just return 3?

buffalu: there's a chance weird case might be hit. let's say we're doing 20 epoch lookback and miss an epoch. everything will be fine until you hit the missing epoch, at which point epoch_range will return nothing, but there's a bunch of data in the buffer

programs/validator-history/src/state.rs
```diff
@@ -454,4 +549,53 @@ mod tests {
...
+            let entry = ValidatorHistoryEntry {
+                epoch: i,
+                ..ValidatorHistoryEntry::default()
+            };
+            circ_buf.push(entry);
+        }
+
+        // Test epoch range [0, 3]
+        let epoch_range = circ_buf.epoch_range(0, 3).unwrap();
+        assert_eq!(epoch_range.len(), 3);
```
buffalu, line 583: can you run assertions these are in the order you expect them to be in?

programs/validator-history/src/state.rs
```diff
@@ -454,4 +549,53 @@ mod tests {
     fn test_validator_history_layout() {
         println!("{}", ValidatorHistoryEntry::type_layout());
     }
+
+    #[test]
+    fn test_epoch_range() {
```
buffalu, line 554: can you add test for wrap-around case?

programs/validator-history/src/state.rs
ebatsell: fixed

programs/validator-history/src/state.rs
ebatsell: returns all existing epochs in range now

programs/validator-history/src/state.rs
ebatsell: done

programs/validator-history/src/state.rs
```diff
@@ -148,6 +178,55 @@ impl CircBuf {
     pub fn arr_mut(&mut self) -> &mut [ValidatorHistoryEntry] {
         &mut self.arr
     }
+
+    /// Returns &ValidatorHistoryEntry for each existing entry in range [start_epoch, end_epoch], factoring for wraparound
+    /// Returns None if either start_epoch or end_epoch is not in the CircBuf
+    pub fn epoch_range(&self, start_epoch: u16, end_epoch: u16) -> Vec<&ValidatorHistoryEntry> {
```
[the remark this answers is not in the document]
ebatsell: we store them as u16 in the history entries, could maybe do the casting inside this fn

Issue #6, opened 2023-12-19
Feature: Add the MEV earned per epoch in the ValidatorHistoryAccount


buffalu:
**Is your feature request related to a problem? Please describe.**
The ValidatorHistoryAccount should track the MEV earned per epoch for every validator on-chain. This allows one to determine the MEV / stake amount on-chain to determine how much MEV relative to stake every validator is receiving.

**Describe the solution you'd like**
The MEV earned every epoch is stored in the [TipDistributionAccount](https://github.com/jito-foundation/jito-programs/blob/28520ea8f3830040452255906732c77a8751e627/mev-programs/programs/tip-distribution/src/state.rs#L32), a PDA owned by the tip distribution program. After the merkle root is uploaded, the [max_total_claim](https://github.com/jito-foundation/jito-programs/blob/28520ea8f3830040452255906732c77a8751e627/mev-programs/programs/tip-distribution/src/state.rs#L62C16-L62C16) field can be referenced to find the total amount of MEV earned by the validator for a given epoch.

This can likely be copied over to the ValidatorHistory account in a modified version of the update_mev_comission instruction, which loads the [tip distribution account](https://github.com/jito-foundation/stakenet/blob/8ce13c7a18822fd65d08ea52a4cce74b6f83f3ef/programs/validator-history/src/instructions/update_mev_commission.rs#L39). There would be a slight change to relax the epoch constraint and pass in the epoch as instruction data and ensure the MEV tips update the ValidatorHistoryAccount for the correct epoch.

In fact, it may be worth it to relax the epoch constraint on the update_mev_comission instruction as well to allow one to backfill missed epochs.

thearyanag:
hey @buffalu , I'd like to work on this, can you please assign me?

PR #13, opened 2024-01-08
feat: add MEV earned per epoch in the ValidatorHistoryAccount

0xprames:
I took a stab at #6 , saw it's been open for a few weeks

mainly out of wanting to learn, this is my first experience w/ Anchor, solana programs, instructions etc etc

I'd love a review to see if I thought about the issue correctly, but i'm happy to close it out or collaborate w/ the assignee of #6 to get this in

### DIFF
--- a/keepers/validator-keeper/src/main.rs
+++ b/keepers/validator-keeper/src/main.rs
@@ -130,8 +130,8 @@ async fn mev_earned_loop(
         // Continuously runs throughout an epoch, polling for tip distribution accounts from the prev epoch with uploaded merkle roots
         // and submitting update_mev_earned (technically update_mev_comission) txs when the uploaded merkle roots are detected
         match update_mev_earned(
-            client.clone(),
-            keypair.clone(),
+            &client,
+            &keypair,
             &commission_history_program_id,
             &tip_distribution_program_id,
             &mut validators_updated,
@@ -399,16 +399,16 @@ async fn main() {
     }
 
     tokio::spawn(mev_commission_loop(
-        Arc::clone(&client),
-        Arc::clone(&keypair),
+        client.clone(),
+        keypair.clone(),
         args.program_id,
         args.tip_distribution_program_id,
         args.interval,
     ));
 
     tokio::spawn(mev_earned_loop(
-        Arc::clone(&client),
-        Arc::clone(&keypair),
+        client.clone(),
+        keypair.clone(),
         args.program_id,
         args.tip_distribution_program_id,
         args.interval,
--- a/keepers/validator-keeper/src/main.rs
+++ b/keepers/validator-keeper/src/main.rs
@@ -18,10 +18,10 @@ use solana_sdk::{
 use tokio::time::sleep;
 use validator_keeper::{
     cluster_info::update_cluster_info,
-    emit_cluster_history_datapoint, emit_mev_commission_datapoint,
+    emit_cluster_history_datapoint, emit_mev_commission_datapoint, emit_mev_earned_datapoint,
     emit_validator_commission_datapoint, emit_validator_history_metrics,
     gossip::{emit_gossip_datapoint, upload_gossip_values},
-    mev_commission::update_mev_commission,
+    mev_commission::{update_mev_commission, update_mev_earned},
     stake::{emit_stake_history_datapoint, update_stake_history},
     vote_account::update_vote_accounts,
 };
@@ -109,6 +109,43 @@ async fn mev_commission_loop(
             Err((e, stats)) => {
                 emit_mev_commission_datapoint(stats);
                 datapoint_error!("mev-commission-error", ("error", e.to_string(), String),);
+                sleep(Duration::from_secs(5)).await;
+            }
+        };
+    }
+}
+
+async fn mev_earned_loop(
+    client: Arc<RpcClient>,
+    keypair: Arc<Keypair>,
+    commission_history_program_id: Pubkey,
+    tip_distribution_program_id: Pubkey,
+    interval: u64,
+) {
+    let mut curr_epoch = 0;
+    // {TipDistributionAccount : VoteAccount}
+    let mut validators_updated: HashMap<Pubkey, Pubkey> = HashMap::new();
+
+    loop {
+        // Continuously runs throughout an epoch, polling for tip distribution accounts from the prev epoch with uploaded merkle roots
+        // and submitting update_mev_earned (technically update_mev_comission) txs when the uploaded merkle roots are detected
+        match update_mev_earned(
+            client.clone(),
+            keypair.clone(),
+            &commission_history_program_id,
+            &tip_distribution_program_id,
+            &mut validators_updated,
+            &mut curr_epoch,
+        )
+        .await
+        {
+            Ok(stats) => {
+                emit_mev_earned_datapoint(stats);
+                sleep(Duration::from_secs(interval)).await;
+            }
+            Err((e, stats)) => {
+                emit_mev_earned_datapoint(stats);
+                datapoint_error!("mev-earned-error", ("error", e.to_string(), String),);
                 sleep(Duration::from_secs(5)).await;
             }
         };
@@ -369,5 +406,12 @@ async fn main() {
         args.interval,
     ));
 
+    tokio::spawn(mev_earned_loop(
+        Arc::clone(&client),
+        Arc::clone(&keypair),
+        args.program_id,
+        args.tip_distribution_program_id,
+        args.interval,
+    ));
     gossip_upload_loop(client, keypair, args.program_id, entrypoint, args.interval).await;
 }
--- a/keepers/validator-keeper/src/mev_commission.rs
+++ b/keepers/validator-keeper/src/mev_commission.rs
@@ -215,7 +215,7 @@ pub async fn update_mev_earned(
     }
     *curr_epoch = epoch;
 
-    let vote_accounts = get_vote_accounts_with_retry(&client, MIN_VOTE_EPOCHS, None)
+    let vote_accounts = get_vote_accounts_with_retry(client, MIN_VOTE_EPOCHS, None)
         .await
         .map_err(|e| (e.into(), CreateUpdateStats::default()))?;
 
@@ -232,7 +232,7 @@ pub async fn update_mev_earned(
         })
         .collect::<Vec<ValidatorMevCommissionEntry>>();
 
-    let uploaded_merkleroot_entries = get_entries_with_uploaded_merkleroot(&client, &entries)
+    let uploaded_merkleroot_entries = get_entries_with_uploaded_merkleroot(client, &entries)
         .await
         .map_err(|e| (e.into(), CreateUpdateStats::default()))?;
 
@@ -241,12 +241,12 @@ pub async fn update_mev_earned(
         .filter(|entry| !validators_updated.contains_key(&entry.tip_distribution_account))
         .collect::<Vec<ValidatorMevCommissionEntry>>();
     let (create_transactions, update_instructions) =
-        build_create_and_update_instructions(&client, &entries_to_update)
+        build_create_and_update_instructions(client, &entries_to_update)
             .await
             .map_err(|e| (e.into(), CreateUpdateStats::default()))?;
 
     let submit_result =
-        submit_create_and_update(&client, create_transactions, update_instructions, &keypair).await;
+        submit_create_and_update(client, create_transactions, update_instructions, keypair).await;
     if submit_result.is_ok() {
         for ValidatorMevCommissionEntry {
             vote_account,
@@ -296,7 +296,7 @@ async fn get_entries_with_uploaded_merkleroot(
         .map(|entry| entry.tip_distribution_account)
         .collect::<Vec<Pubkey>>();
 
-    let accounts = get_multiple_accounts_batched(&tip_distribution_addresses, &client).await?;
+    let accounts = get_multiple_accounts_batched(&tip_distribution_addresses, client).await?;
     let result = accounts
         .iter()
         .enumerate()
--- a/keepers/validator-keeper/src/mev_commission.rs
+++ b/keepers/validator-keeper/src/mev_commission.rs
@@ -34,6 +34,7 @@ pub struct ValidatorMevCommissionEntry {
     pub config: Pubkey,
     pub program_id: Pubkey,
     pub signer: Pubkey,
+    pub epoch: u64,
 }
 
 impl ValidatorMevCommissionEntry {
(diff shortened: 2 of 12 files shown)


programs/validator-history/src/state.rs
```diff
@@ -185,7 +189,12 @@ impl ValidatorHistory {
...
     pub const SEED: &'static [u8] = b"validator-history";
 
-    pub fn set_mev_commission(&mut self, epoch: u16, commission: u16) -> Result<()> {
+    pub fn set_mev_commission(
+        &mut self,
+        epoch: u16,
+        commission: u16,
+        mev_earned: u64,
+    ) -> Result<()> {
         // check if entry exists for the epoch
```
buffalu, line 198: this logic is wrong. mentioned to @ebatsell in diff PR, but if you update epoch 500, then 501, then 500 you'll end up with two epoch 500 entries. the second 500 update should go back and find the first 500 entry

0xprames: yeah for sure I see the other PR

Im thinking if calling [self.history.epoch_range(epoch,epoch)](https://github.com/jito-foundation/stakenet/pull/12/files#diff-d658c99b6787b63bdf8cc47cb1dab02627e07efad1ccb54b44bc4898302c336eR184 ) from that PR here on the circbuf will fix the logic

0xprames: nvm didnt need that, i pushed up what i think should address it

i think the case you mention would fail the existing check bc the `last_mut()`.epoch would != 500

```
  if let Some(entry) = self.history.last_mut() {
            if entry.epoch == epoch { // i don't think this would be true
                entry.mev_commission = commission;}
  
```
but yeah, a subequent update would fail, so the entry data could be stale (?)

0xprames: i deleted my prev comment on this thread - since it was incorrect

you're right, with the existing logic, in the case where we try and edit an epoch seen prev, it will create another entry for it.

i went ahead and pushed up a fix for that

ebatsell: high level:
* the CircBuf struct should have at most one entry per epoch, and they should be laid out in ascending order
* if for some reason an entry is missed, it's pretty messy to try to insert after the fact due to the need to shift every following epoch down by one. 
* a new entry should be appended if an entry doesn't exist for that epoch and it's greater than the most recent epoch (`.last()`)

therefore the update pseudocode for all of these `set_`s should be something like:
```
if epoch > circ_buf.last().epoch {
  insert new entry
} else {
   if let Some(entry) = circ_buf.find(epoch) {
     modify entry
   }
}
```

keepers/validator-keeper/src/mev_commission.rs
```diff
@@ -122,7 +124,7 @@ impl UpdateInstruction for ValidatorMevCommissionEntry {
                 signer: self.signer,
             }
             .to_account_metas(None),
-            data: validator_history::instruction::UpdateMevCommission {}.data(),
+            data: validator_history::instruction::UpdateMevCommission { epoch: self.epoch }.data(),
```
ebatsell, line 127: Merkle roots for a given epoch are uploaded typically within a few hours after the epoch ends, so to account for this, the keeper will need to submit UpdateMevCommission instructions for the previous epoch after its upload has completed. could achieve this by adding a separate loop in the `main` that submits these at say, the halfway point of the next epoch. Or better yet, continually checks the tip distribution accounts to see which merkle_roots are not None

programs/validator-history/src/state.rs
```diff
@@ -52,15 +52,18 @@ pub struct ValidatorHistoryEntry {
...
     pub padding0: u8,
     // 0 if not a superminority validator, 1 if superminority validator
     pub is_superminority: u8,
     // rank of validator by stake amount
     pub rank: u32,
     // Most recent updated slot for epoch credits and commission
     pub vote_account_last_update_slot: u64,
-    pub padding1: [u8; 88],
+    // MEV earned
+    pub mev_earned: u64,
```
ebatsell, line 65: in the interest of using the remaining space here wisely, I think it'd be best to represent this as a [fixed point decimal](https://en.wikipedia.org/wiki/Fixed-point_arithmetic). if this is 4 bytes with 2 decimals of precision that gives us a max MEV earned of 42949672.95 SOL for 1 validator for 1 epoch which won't happen for.. a long time 😄

programs/validator-history/src/state.rs
0xprames: yeah makes sense, should be taken care of for all the `set_`s now

keepers/validator-keeper/src/mev_commission.rs
0xprames: for sure, lmk if what i have works

i added a `update_mev_earned` loop similar to the existing `update_mev_comission` with the main difference that instead of just looking for existing TDAs, it attempts to `TipDistributionAccount::try_deserialize` on the `account.data.data` received from the gMA call and and check for `Some(merkle_root) = tda.merkle_root`

programs/validator-history/src/state.rs
0xprames: for sure, lmk if what i pushed up is correct

the internal u32 representation / 100 should give the max mev earned in sol with 2 dec of precision

keepers/validator-keeper/src/mev_commission.rs
```diff
@@ -215,3 +285,38 @@ async fn get_existing_entries(
...
+                        return Some(entries[i].clone());
+                    } else {
+                        return None;
+                    }
+                } else {
+                    return None;
+                }
+            } else {
+                return None;
+            }
```
ebatsell, line 317: nit: could get rid of all the `else`s and just set the last line as `None`

ebatsell (commented):
Since we're doing more than just setting mev_commission in this instruction, I think we should also rename the instruction to `copy_tip_distribution_account`

programs/validator-history/src/instructions/update_mev_commission.rs
```diff
@@ -48,16 +47,20 @@ pub struct UpdateMevCommission<'info> {
...
 
-pub fn handler(ctx: Context<UpdateMevCommission>) -> Result<()> {
+pub fn handler(ctx: Context<UpdateMevCommission>, epoch: u64) -> Result<()> {
     let mut validator_history_account = ctx.accounts.validator_history_account.load_mut()?;
 
     let mut tda_data: &[u8] = &ctx.accounts.tip_distribution_account.try_borrow_data()?;
 
     let tip_distribution_account = TipDistributionAccount::try_deserialize(&mut tda_data)?;
     let mev_commission_bps = tip_distribution_account.validator_commission_bps;
-    let epoch = cast_epoch(Clock::get()?.epoch);
```
ebatsell, line 58: We need to do a check here and anywhere else `epoch` is passed in that `epoch` is not greater than `Clock::get().epoch`

keepers/validator-keeper/src/mev_commission.rs
ebatsell: perfect

programs/validator-history/src/state.rs
```diff
@@ -369,10 +393,24 @@ impl ValidatorHistory {
...
+                    entry.commission = commission;
+                    entry.vote_account_last_update_slot = slot;
+                    return Ok(());
+                }
+                Ordering::Greater => {
+                    self.history
+                        .arr_mut()
+                        .iter_mut()
+                        .filter(|entry| entry.epoch == epoch)
+                        .for_each(|entry| {
```
ebatsell, line 407: this implies there may be multiple entries with the same epoch in the CircBuf, which shouldn't be possible given how we're inserting new entries - would prefer to use `.find(|entry| entry.epoch == epoch)` here instead of `.filter().for_each(..)`

0xprames: makes sense

programs/validator-history/src/instructions/update_mev_commission.rs
0xprames: done, the other 2 instructions (update_stake_history and backfill_total_blocks) seem to have the check

keepers/validator-keeper/src/mev_commission.rs
0xprames: yeah good call, done

programs/validator-history/src/utils.rs
```diff
@@ -4,6 +4,16 @@ pub fn cast_epoch(epoch: u64) -> u16 {
     (epoch % u16::MAX as u64).try_into().unwrap()
 }
 
+pub fn fixed_point_sol(lamports: u64) -> u32 {
```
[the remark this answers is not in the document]
0xprames: done

programs/validator-history/src/utils.rs
```diff
@@ -4,6 +4,16 @@ pub fn cast_epoch(epoch: u64) -> u16 {
...
+    // convert to sol
+    let mut sol = lamports as f64 / 1000000000.00;
```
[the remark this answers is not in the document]
0xprames: done

programs/validator-history/src/state.rs
```diff
@@ -52,15 +52,18 @@ pub struct ValidatorHistoryEntry {
...
+    // without it - `rank` would have offset 27, and the compiler would add an implicit padding byte after `is_superminority` and before `rank`
     pub padding0: u8,
     // 0 if not a superminority validator, 1 if superminority validator
     pub is_superminority: u8,
     // rank of validator by stake amount
     pub rank: u32,
     // Most recent updated slot for epoch credits and commission
     pub vote_account_last_update_slot: u64,
-    pub padding1: [u8; 88],
+    // MEV earned
```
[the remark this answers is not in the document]
0xprames: done

0xprames:
> Since we're doing more than just setting mev_commission in this instruction, I think we should also rename the instruction to copy_tip_distribution_account

done - `UpdateMevComission` -> `CopyTipDistributionAccount`

ebatsell:
@0xprames looks great, once all tests and checks are passing I'll approve

keepers/validator-keeper/src/main.rs
```diff
@@ -369,5 +406,12 @@ async fn main() {
         args.interval,
     ));
 
+    tokio::spawn(mev_earned_loop(
+        Arc::clone(&client),
```
buffalu, line 410: this can be client.clone()

keepers/validator-keeper/src/mev_commission.rs
```diff
@@ -190,6 +194,73 @@ pub async fn update_mev_commission(
...
+    let vote_accounts = get_vote_accounts_with_retry(&client, MIN_VOTE_EPOCHS, None)
+        .await
+        .map_err(|e| (e.into(), CreateUpdateStats::default()))?;
+
+    let entries = vote_accounts
+        .iter()
+        .map(|vote_account| {
+            ValidatorMevCommissionEntry::new(
+                vote_account,
+                epoch - 1, // TDA derived from the prev epoch since the merkle roots are uploaded shortly after rollover
```
buffalu, line 227: ultra nit: saturating_sub for epoch 0

keepers/validator-keeper/src/mev_commission.rs
```diff
@@ -190,6 +194,73 @@ pub async fn update_mev_commission(
...
+                validator_history_program_id,
+                tip_distribution_program_id,
+                &keypair.pubkey(),
+            )
+        })
+        .collect::<Vec<ValidatorMevCommissionEntry>>();
+
+    let uploaded_merkleroot_entries =
+        get_entries_with_uploaded_merkleroot(client.clone(), &entries)
```
buffalu, line 236: use &client

keepers/validator-keeper/src/mev_commission.rs
```diff
@@ -215,3 +286,33 @@ async fn get_existing_entries(
     // Fetch existing tip distribution accounts for this epoch
     Ok(result)
 }
+
+async fn get_entries_with_uploaded_merkleroot(
+    client: Arc<RpcClient>,
```
buffalu, line 291: use &Arc<RpcClient>

keepers/validator-keeper/src/mev_commission.rs
```diff
@@ -215,3 +286,33 @@ async fn get_existing_entries(
...
+        .map(|entry| entry.tip_distribution_account)
+        .collect::<Vec<Pubkey>>();
+
+    let accounts = get_multiple_accounts_batched(&tip_distribution_addresses, &client).await?;
+    let result = accounts
+        .iter()
+        .enumerate()
+        .filter_map(|(i, account_data)| {
+            if let Some(account_data) = account_data {
+                let mut data: &[u8] = &account_data.data;
```
buffalu, line 306: nit: can remove the if let Some() and do something like this:
```
let tda = TipDistributionAccount::try_deserialize(&mut account_data?.data.as_slice()).ok()?;
if tda.merkle_root.is_some() { ...} else { None }

File: programs/validator-history/src/instructions/copy_tip_distribution_account.rs [hunk 15]
```diff
@@ -48,16 +49,25 @@ pub struct UpdateMevCommission<'info> {
     pub signer: Signer<'info>,
 }
 
-pub fn handler(ctx: Context<UpdateMevCommission>) -> Result<()> {
+pub fn handler(ctx: Context<CopyTipDistributionAccount>, epoch: u64) -> Result<()> {
+    // Cannot set stake for future epochs
```
buffalu, line 53: comment?

buffalu (commented):
mostly stylistic comments!

keepers/validator-keeper/src/main.rs
0xprames: done, modified in the existing code as well

keepers/validator-keeper/src/mev_commission.rs
0xprames: makes sense

keepers/validator-keeper/src/mev_commission.rs
0xprames: `account_data?` didnt seem to work bc its `Option<>`,  i left this as for now

programs/validator-history/src/instructions/copy_tip_distribution_account.rs
0xprames: good catch

keepers/validator-keeper/src/mev_commission.rs
```diff
@@ -190,6 +193,73 @@ pub async fn update_mev_commission(
     submit_result.map_err(|(e, stats)| (e.into(), stats))
 }
 
+pub async fn update_mev_earned(
+    client: Arc<RpcClient>,
```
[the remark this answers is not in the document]
0xprames: makes sense, i def should borrow more than move into when i can

ebatsell (approved):

PR #19, opened 2024-01-25
Improve Keeper Reliability

ebatsell:
Improves keeper reliability for landing transactions by:
* no longer skipping transactions that initially fail with a BlockhashNotFound error
* Merging vote accounts with all validator history accounts for epoch credit cranking, so offline validators that no longer showing up in getVoteAccounts still get cranked (necessary for Steward program scoring)

The first part has already been running on mainnet for a week, and for the last 3 epochs, all vote accounts have been updated each epoch (as measured by: same number of commissions tracked as stakes).

### DIFF
--- a/keepers/keeper-core/src/lib.rs
+++ b/keepers/keeper-core/src/lib.rs
@@ -1,11 +1,9 @@
-use std::borrow::BorrowMut;
 use std::fmt::{Display, Formatter};
 use std::mem::size_of;
-use std::sync::Mutex;
 use std::vec;
 use std::{collections::HashMap, sync::Arc, time::Duration};
 
-use clap::{error, ValueEnum};
+use clap::ValueEnum;
 use log::*;
 use solana_client::rpc_response::RpcVoteAccountInfo;
 use solana_client::{client_error::ClientError, nonblocking::rpc_client::RpcClient};
@@ -224,7 +222,7 @@ async fn parallel_submit_transactions(
             )
         })?;
         // Convert instructions to transactions in batches and send them all, saving their signatures
-        let transactions: Vec<Transaction> = transaction_batch
+        let submitted_transactions: Vec<Transaction> = transaction_batch
             .iter()
             .map(|batch| {
                 Transaction::new_signed_with_payer(
@@ -236,7 +234,7 @@ async fn parallel_submit_transactions(
             })
             .collect();
 
-        let tx_futures = transactions
+        let tx_futures = submitted_transactions
             .iter()
             .map(|tx| async move { client.send_transaction(tx).await })
             .collect::<Vec<_>>();
@@ -251,15 +249,13 @@ async fn parallel_submit_transactions(
                     match e.get_transaction_error() {
                         // If blockhash not found, transaction is still valid and should not be dropped
                         Some(TransactionError::BlockhashNotFound) => {
-                            executed_signatures.insert(
-                                transactions[i + index_offset].signatures[0],
-                                i + index_offset,
-                            );
+                            executed_signatures
+                                .insert(submitted_transactions[i].signatures[0], i + index_offset);
                         }
                         // If another error is returned, transaction probably won't succeed on retries
                         Some(_) | None => {
                             let message = e.to_string();
-                            warn!("Transaction failed: {}", message);
+                            warn!("Transaction failed: {:?}", e);
                             error_messages.insert(i + index_offset, message);
                         }
                     }
@@ -269,7 +265,7 @@ async fn parallel_submit_transactions(
 
         debug!(
             "Transactions sent: {}, executed_signatures: {}",
-            transactions.len(),
+            submitted_transactions.len(),
             executed_signatures.len()
         );
     }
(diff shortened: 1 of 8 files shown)


keepers/keeper-core/src/lib.rs
```diff
@@ -237,8 +238,17 @@ async fn parallel_submit_transactions(
...
                     Ok(signature) => Some((signature, i)),
                     Err(e) => {
-                        warn!("Send transaction failed: {:?}", e);
-                        None
+                        match e.get_transaction_error() {
+                            // If blockhash not found, transaction is still valid and should not be dropped
+                            Some(TransactionError::BlockhashNotFound) => {
+                                Some((tx.signatures[0], i))
+                            }
+                            // If another error is returned, transaction probably won't succeed on retries
```
buffalu, line 246: i think we want to know if this is happening

keepers/keeper-core/src/lib.rs
```diff
@@ -298,107 +248,135 @@ async fn parallel_confirm_transactions(
...
+            break;
         }
 
-        // Update instructions to the ones remaining
-        instructions = executed_signatures
-            .into_values()
-            .flat_map(|i| instruction_batches[i])
-            .cloned()
-            .collect::<Vec<_>>();
+        // Re-sign transactions with fresh blockhash
```
buffalu, line 335: going to send duplicates if always resigning. should just need to resign once the blockhash you previously signed txs with expires based on processed commit level. there's a is_blockhash_valid_with_config method

keepers/keeper-core/src/lib.rs
```diff
@@ -298,107 +248,135 @@ async fn parallel_confirm_transactions(
...
-        let instruction_batches: Vec<&[Instruction]> =
-            instructions.chunks(instructions_per_tx).collect();
-        parallel_submit_transactions(
-            client,
-            signer,
-            &instruction_batches,
-            &mut executed_signatures,
-        )
-        .await?;
+    while retries < retry_count {
```
buffalu, line 296: imo max retries should prolly be incremented on blockhash increment? or related to some time.

retry_count of 3 may be fine for 100 txs, but won't be fine for 2k txs

keepers/keeper-core/src/lib.rs
```diff
@@ -30,14 +33,20 @@ pub struct CreateUpdateStats {
...
-    #[error("Transactions failed to execute after multiple retries")]
-    RetryError(Vec<Instruction>),
-    #[error("Transactions failed to execute after multiple retries")]
-    TransactionRetryError(Vec<Vec<Instruction>>),
     #[error("RPC Client error: {0:?}")]
-    ClientError(String, Vec<Instruction>),
+    ClientError(String),
     #[error("RPC Client error: {0:?}")]
-    TransactionClientError(String, Vec<Vec<Instruction>>),
+    TransactionClientError(String, Vec<Result<(), SendTransactionError>>),
```
buffalu, line 39: this is kinda messy return type

buffalu: what does vec correspond to?

keepers/keeper-core/src/lib.rs
```diff
@@ -298,107 +248,135 @@ async fn parallel_confirm_transactions(
     info!(
         "{} transactions submitted, {} confirmed",
         num_transactions_submitted,
-        num_transactions_submitted - executed_signatures.len()
+        confirmed_signatures.len()
     );
+    confirmed_signatures.clone()
```
buffalu, line 253: shouldn't need to clone

keepers/keeper-core/src/lib.rs
```diff
@@ -31,15 +34,15 @@ pub type Error = Box<dyn std::error::Error>;
 #[derive(ThisError, Debug, Clone)]
 pub enum TransactionExecutionError {
     #[error("Transactions failed to execute after multiple retries")]
-    RetryError(Vec<Instruction>),
-    #[error("Transactions failed to execute after multiple retries")]
-    TransactionRetryError(Vec<Vec<Instruction>>),
+    TransactionRetryError(Vec<(Vec<Instruction>, String)>),
```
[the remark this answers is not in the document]
ebatsell: there's other transaction types baked into the ClientError like IoError, ReqwestError, SigningError, etc that I feel is useful to propogate. and the only way to bubble it up is to cast to string since the ClientError type doesn't implement the Clone trait we need. If we saved only the transaction errors you'd end up with `Vec<Result<(), Option<TransactionError>>>`. Can switch that out if it's preferred but we'd be losing information with that

keepers/keeper-core/src/lib.rs
```diff
@@ -197,71 +206,11 @@ async fn calculate_instructions_per_tx(
...
-    }
-    Ok(())
-}
-
 async fn parallel_confirm_transactions(
     client: &RpcClient,
-    executed_signatures: &mut HashMap<Signature, usize>,
-) {
-    // Confirmes TXs in batches of 256 (max allowed by RPC method)
+    executed_signatures: HashMap<Signature, usize>,
```
buffalu, line 211: nit: you could pass in hashset here then return hashset of confirmed and worry about the index in caller. but this is fine

keepers/keeper-core/src/lib.rs
ebatsell: using a retry count of 10, maybe you were looking at a older commit?

keepers/keeper-core/src/lib.rs
ebatsell: done

keepers/keeper-core/src/lib.rs
ebatsell: added logic to fetch blockhash and increment retries only when it's expired

keepers/keeper-core/src/lib.rs
ebatsell: added logic to do less re-signing

buffalu (approved):

PR #20, opened 2024-02-08
Run all program-test in SBF mode and speed up CI

buffalu:
- Run in SBF mode to make sure actual BPF and runtime restrictions are being enforced. This should help us catch runtime errors before they make it to prod.
- Speed up CI to get fast iteration cycles
- Cleanup some compiler warnings by renaming functions to not have overlap

Screenshot 2024-02-07 at 11 48 27 PM


.github/workflows/build.yaml
```diff
@@ -3,97 +3,149 @@ on:
...
+      - uses: actions-rust-lang/setup-rust-toolchain@v1
+      - name: Install solana-verify from crates.io
+        uses: baptiste0928/cargo-install@v3
+        with:
+          crate: solana-verify
+      - name: Install anchor-cli from crates.io
+        uses: baptiste0928/cargo-install@v3
+        with:
+          crate: anchor-cli
+          version: "0.29.0"
```
ebatsell, line 77: this should be anchor 0.28

programs/validator-history/idl/validator_history.json
```diff
@@ -797,7 +797,7 @@
                 "type": "u8"
               },
               {
-                "name": "num_addrs",
+                "name": "numAddrs",
```
ebatsell, line 800: this shouldn't change as long as we're still on anchor 0.28, maybe the solana program test compatibility is fixed with the new way we're loading in programs but we'd still need to bump versions and make the code compatible

ebatsell (commented):
so what's the new command to run tests? the `test` script you had earlier is missing

.github/workflows/build.yaml
buffalu: CI not liking 0.28.0. unless you feel strongly about it, 0.29.0 seems fine as 0.28.0 never made it to crates.io

buffalu:
> so what's the new command to run tests? the `test` script you had earlier is missing

added to readme

ebatsell (approved):

PR #24, opened 2024-03-04
Miscellaneous fixes and cleanup

ebatsell:
Fixes a number of small things including:

* Potential for self invocation of CopyGossipContactInfo to overwrite real values
* Reading data offsets in CopyGossipContactInfo rather than hardcoding assumed offsets
* Potential epoch overflow for cast epoch
* Comments around how specific structs and instructions are intended to work
* Shifting epoch_start_timestamp to avoid overwriting existing bytes and avoid trimming data type


programs/validator-history/src/instructions/copy_gossip_contact_info.rs
```diff
@@ -2,12 +2,34 @@ use anchor_lang::{
...
-    crds_value::CrdsData, errors::ValidatorHistoryError, state::ValidatorHistory, utils::cast_epoch,
+    crds_value::CrdsData, errors::ValidatorHistoryError, state::ValidatorHistory,
+    utils::cast_epoch, Config,
 };
 use validator_history_vote_state::VoteStateVersions;
 
+// Structs and constants copied from solana_sdk::ed25519_instruction. Copied in order to make fields public. Compilation issues hit when importing solana_sdk
+#[derive(Default, Debug, Copy, Clone, Zeroable, Pod, Eq, PartialEq)]
+#[repr(C)]
+pub struct Ed25519SignatureOffsets {
```
buffalu, line 16: why can't import thes from solana_sdk?

buffalu: nvmd

programs/validator-history/src/utils.rs
```diff
@@ -1,13 +1,21 @@
...
+    prelude::{AccountInfo, Pubkey, Result},
+    require,
     solana_program::native_token::lamports_to_sol,
 };
 
-pub fn cast_epoch(epoch: u64) -> u16 {
-    (epoch % u16::MAX as u64).try_into().unwrap()
+use crate::errors::ValidatorHistoryError;
+
+pub fn cast_epoch(epoch: u64) -> Result<u16> {
```
buffalu, line 9: result not used. require panics + unwrap too.

programs/validator-history/src/instructions/copy_gossip_contact_info.rs
ebatsell: Running into some bad version conflicts on inner packages of solana-sdk when mixing it with anchor. have hit this before with things like the gossip structs and we ended up copying those over as well

programs/validator-history/src/utils.rs
ebatsell: `require` returns an `Err` if the condition is hit, so the anchor Result is needed here

programs/validator-history/src/utils.rs
```diff
@@ -1,13 +1,21 @@
...
+    require!(
+        epoch < (u16::MAX as u64),
+        ValidatorHistoryError::EpochTooLarge
+    );
+    let epoch_u16: u16 = (epoch % u16::MAX as u64).try_into().unwrap();
```
buffalu, line 14: nit: can't you do try_into()? unwrap in fn with Result<>

programs/validator-history/src/utils.rs
ebatsell: I think it's a macro that expands into something that returns Err so the ? is not needed. If you have require in a function that doesn't return Result<> it complains

programs/validator-history/src/utils.rs
ebatsell: the Result here is secretly Result<T, AnchorError> which is incompatible with the try_into return type here but since we know it's gonna be a valid u16 from the line above the unwrap is fine

buffalu (approved):

PR #37, opened 2024-05-23
Insert missing entries in CopyVoteAccount

ebatsell:
Problem: 
Due to unreliability in landing transactions in epoch 598 (when chain was super congested), about half of active validators did not get a single ValidatorHistoryEntry created, so their epoch credits were never copied over in future epochs. This is bad for Stakenet scoring, and this issue could happen in the future if congestion gets bad again.

Solution:
In the CopyVoteAccount instruction, when epochs are detected in the vote account that don't exist in the CircBuf, insert those new entries and shift all entries with greater epochs forward by one. This will evict the oldest entries if we are wrapping around. After that, epoch credits are copied to the old entries. We will then be able to retroactively call:
`CopyTipDistributionAccount` and `UpdateStakeHistory` on those validators. 

Notes:
* We are still within the 64 epoch vote credit range, so epoch credits from 598 and any other randomly skipped entries will get added
* In the most extreme case possible of 64 entries inserted (max epochs in vote account), this instruction uses ~270k CUs. Most of the time none will be added, and when this first runs, a handful will be added for some validators. Normal CU usage is <30k.

### DIFF
--- a/programs/validator-history/src/state.rs
+++ b/programs/validator-history/src/state.rs
@@ -442,12 +442,13 @@ impl ValidatorHistory {
             }));
 
         for (entry_is_some, epoch) in entries.iter().zip(start_epoch as u16..=end_epoch) {
-            if !*entry_is_some && epoch_credits_map.get(&epoch).is_some() {
+            if !*entry_is_some && epoch_credits_map.contains_key(&epoch) {
+                // Inserts blank entry that will have credits copied to it later
                 let entry = ValidatorHistoryEntry {
                     epoch,
                     ..ValidatorHistoryEntry::default()
                 };
-                // If entry cannot be inserted, skips
+                // Skips if epoch is out of range or duplicate
                 self.history.insert(entry, epoch).unwrap_or_default();
             }
         }
--- a/programs/validator-history/src/state.rs
+++ b/programs/validator-history/src/state.rs
@@ -4,7 +4,7 @@ use {
         errors::ValidatorHistoryError,
         utils::cast_epoch,
     },
-    anchor_lang::prelude::*,
+    anchor_lang::{prelude::*, solana_program::log::sol_log_compute_units},
     borsh::{BorshDeserialize, BorshSerialize},
     std::{cmp::Ordering, collections::HashMap, mem::size_of, net::IpAddr},
     type_layout::TypeLayout,
@@ -37,7 +37,7 @@ impl Config {
 
 static_assertions::const_assert_eq!(size_of::<ValidatorHistoryEntry>(), 128);
 
-#[derive(AnchorSerialize, TypeLayout)]
+#[derive(AnchorSerialize, TypeLayout, Debug)]
 #[zero_copy]
 pub struct ValidatorHistoryEntry {
     pub activated_stake_lamports: u64,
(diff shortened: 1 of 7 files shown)


programs/validator-history/src/state.rs
```diff
@@ -184,8 +184,53 @@ impl CircBuf {
...
-    /// Returns None if either start_epoch or end_epoch is not in the CircBuf
+    /// Given a new entry and epoch, inserts the entry into the buffer in sorted order
+    /// Will not insert if the epoch is out of range or already exists in the buffer
+    fn insert(&mut self, entry: ValidatorHistoryEntry, epoch: u16) -> Result<()> {
+        if self.is_empty() {
+            return Err(ValidatorHistoryError::EpochOutOfRange.into());
+        }
+
+        // Find the lowest epoch in the buffer to ensure the new epoch is valid
+        let min_epoch = {
```
CoachChuckFF, line 195: Consider putting this `get_min_epoch` and get_max_epoch` in a helper function

programs/validator-history/src/utils.rs
```diff
@@ -36,6 +36,55 @@ pub fn get_vote_account(validator_history_account_info: &AccountInfo) -> Pubkey
...
+/// Finds the position to insert a new entry with the given epoch, where the epoch is greater than the previous entry and less than the next entry.
+/// Assumes entries are in sorted order (according to CircBuf ordering), and there are no duplicate epochs.
+pub fn find_insert_position(
+    arr: &[ValidatorHistoryEntry],
+    idx: usize,
+    epoch: u16,
+) -> Option<usize> {
+    let len = arr.len();
+    if len == 0 {
+        return None;
```
CoachChuckFF, line 48: Wouldn't an empty array return `Some(0)`?

CoachChuckFF (commented):
Everything looks good!

Two suggestions:
1. Consider making `get_max_epoch` and `get_min_epoch` as helper functions in the Circle Buf
2. Potentially add another integration test(s) in `test_vote_account` to check for wraparounds

programs/validator-history/src/state.rs
```diff
@@ -364,6 +409,65 @@ impl ValidatorHistory {
...
+
+        let end_epoch = cast_epoch(
+            epoch_credits
+                .iter()
+                .max_by_key(|(epoch, _, _)| *epoch)
+                .unwrap()
+                .0,
+        )?;
+
+        let entries = self
```
buffalu, line 440: idk if this collect is needed, can you iter/enumerate in the for loop below?

programs/validator-history/src/state.rs
```diff
@@ -184,8 +184,53 @@ impl CircBuf {
         &mut self.arr
     }
 
-    /// Returns &ValidatorHistoryEntry for each existing entry in range [start_epoch, end_epoch], factoring for wraparound
-    /// Returns None if either start_epoch or end_epoch is not in the CircBuf
+    /// Given a new entry and epoch, inserts the entry into the buffer in sorted order
+    /// Will not insert if the epoch is out of range or already exists in the buffer
+    fn insert(&mut self, entry: ValidatorHistoryEntry, epoch: u16) -> Result<()> {
```
buffalu, line 189: rip, not doing index = epoch % self.history.len() might have been a mistake

ebatsell: 💯

programs/validator-history/src/state.rs
ebatsell: forgot - it's actually necessary cause this is an immutable borrow of self.history and then we mutably borrow when inserting the missing entries, so need to make a copy. will be fine cause it's a vec of bools which gets allocated on heap

PR #39, opened 2024-05-28
FIX: fixed not firing and doing all vote accounts

coachchucksol:
- The copy vote account will now be run at 10%, 50%, and 90% of the epoch ( it was at 0.1% before )
- The copy vote account now updates on all voting accounts less the closes ones, before it was only calling it on all of the active vote accounts


ebatsell (approved):
lgtm!

PR #42, opened 2024-06-03
Add historical_commission_threshold and fixes instant unstaking bug

ebatsell:
* Adds `historical_commission_threshold` as described in the Steward README. Requires a compiler feature when building the main image to specify VALIDATOR_HISTORY_FIRST_RELIABLE_EPOCH for the given network this is running on.
* Fixes a bug in which validators with stake that are not in the delegation set would not be checked for instant unstake status, which was wrong because we still need to be able to instant unstake these validators if they commission rug or go delinquent. 
* Re-enables testing of Steward tests in CI.


coachchucksol (changes requested):
All looks good, just add the new parameter to the readme

coachchucksol (approved):
All looks good to me

PR #43, opened 2024-06-03
FIX + TESTS (STEWARD): Update Parameters

coachchucksol:
**FIXES (STEWARD):**
- `initalize_config` now takes `UpdateParametersArgs` as an argument to ensure the config is in a valid state on inception
- fixed `update_parameters` such that the updated parameters persist ( before they did not )

**TESTS (STEWARD):**
- ➕ `auto_remove_validator`
- ➕ `test_update_parameters`


tests/tests/steward/test_steward.rs
```diff
@@ -121,6 +100,105 @@ async fn test_auto_add_validator_to_pool() {
...
+        &validator_history::id(),
+    );
+
+    let (stake_account_address, transient_stake_account_address, withdraw_authority) =
+        fixture.stake_accounts_for_validator(vote_account).await;
+
+    // Add vote account
+    println!("Adding vote account");
+    _auto_add_validator_to_pool(&fixture, &vote_account).await;
+    println!("Done adding vote account");
```
ebatsell, line 139: nit: remove prints

tests/tests/steward/test_steward.rs
```diff
@@ -303,3 +381,221 @@ async fn test_set_new_authority() {
...
+    );
+
+    fixture.submit_transaction_assert_success(tx).await;
+
+    let config: Config = fixture
+        .load_and_deserialize(&fixture.steward_config.pubkey())
+        .await;
+
+    if let Some(mev_commission_range) = update_parameters.mev_commission_range {
+        assert!(
```
ebatsell, line 413: nit: prefer `assert_eq!`

ebatsell (commented):
could we also get a basic test to check that sending parameters in the init config works?

ebatsell (approved):
I think the default parameters on config init are causing `test_compute_instant_unstake` to fail, should be able to fix by setting `instant_unstake_epoch_progress` lower in that test

PR #47, opened 2024-06-11
Christian/epoch update

coachchucksol:
Starting to add the solution for keeping track of removed validators


programs/steward/src/state/steward_state.rs
```diff
@@ -467,7 +508,40 @@ impl StewardState {
...
+        {
+            // Skip if already processed
+            if self.progress.get(index)? {
+                return Ok(());
+            }
+        }
+
+        {
+            // Skip if not active
+            if stake_status != StakeStatus::Active {
```
ebatsell, line 522:
```suggestion
            if !matches!(stake_status, StakeStatus::Active) {
```

programs/steward/src/instructions/epoch_update.rs
```diff
@@ -0,0 +1,28 @@
...
+        seeds = [StewardStateAccount::SEED, config.key().as_ref()],
+        bump
+    )]
+    pub state_account: AccountLoader<'info, StewardStateAccount>,
+}
+
+// Removes validator from blacklist. Validator will be eligible to receive delegation again when scores are recomputed.
+// Index is the index of the validator from ValidatorHistory.
+pub fn handler(ctx: Context<EpochUpdate>) -> Result<()> {
+    let mut state_account = ctx.accounts.state_account.load_mut()?;
```
ebatsell, line 19: should have a check here on the StakePool account that last_update_epoch == clock.epoch

programs/steward/src/state/steward_state.rs
```diff
@@ -39,8 +42,13 @@ pub fn maybe_transition_and_emit(
     params: &Parameters,
     epoch_schedule: &EpochSchedule,
 ) -> Result<()> {
+    if clock.epoch != state_account.current_epoch {
```
ebatsell, line 45: Better to have this as a `require!()` at the beginning of all state machine instructions rather than hit this function and fail at the end, less CUs used and more clear

programs/steward/src/state/steward_state.rs
```diff
@@ -451,6 +455,43 @@ impl StewardState {
         Ok(())
     }
 
+    /// Marks a validator as skipped in the current state
+    pub fn skip_bad_validator(&mut self, index: usize) -> Result<()> {
```
ebatsell, line 459: not called anywhere

programs/steward/src/state/steward_state.rs
coachchucksol: Done

programs/steward/src/instructions/epoch_update.rs
coachchucksol: Done

programs/steward/src/state/steward_state.rs
coachchucksol: Done

programs/steward/src/state/steward_state.rs
coachchucksol: Removed

programs/steward/src/state/steward_state.rs
```diff
@@ -447,7 +471,26 @@ impl StewardState {
...
+    }
+
+    /// Called when adding a validator to the pool so that we can ensure a 1-1 mapping between
+    /// the validator list and the steward state
+    pub fn increment_validator_to_add(&mut self) -> Result<()> {
+        self.validators_added = self
+            .validators_added
+            .checked_add(1)
+            .ok_or(StewardError::ArithmeticError)?;
         Ok(())
```
ebatsell, line 494: nit: can just be the statement without the Ok(())

programs/steward/src/state/steward_state.rs
```diff
@@ -447,7 +471,26 @@ impl StewardState {
         self.delegations[self.num_pool_validators] = Delegation::default();
         self.instant_unstake.set(self.num_pool_validators, false)?;
         self.progress.set(self.num_pool_validators, false)?;
+        self.progress.set(self.num_pool_validators, false)?;
```
ebatsell, line 474: duplicate line

programs/steward/src/state/steward_state.rs
```diff
@@ -115,11 +123,24 @@ pub struct StewardState {
...
 
+    pub _padding0: [u8; STATE_PADDING_0_SIZE],
+
+    /// Number of validators added to the pool in the current cycle
+    pub validators_added: u16,
+
+    // TODO add closer to the top
+    /// Marks a validator for removal after `remove_validator_from_pool` has been called on the stake pool
+    /// This is cleaned up in the next epoch
+    pub validators_to_remove: BitMask,
```
ebatsell, line 134: Best to move this up now imo since we'll need to do a state reset anyway, the validators_added could go before or after the U8Bools

programs/steward/src/state/steward_state.rs
```diff
@@ -34,26 +34,34 @@ pub struct StateTransition {
...
     params: &Parameters,
     epoch_schedule: &EpochSchedule,
 ) -> Result<()> {
-    let initial_state = state_account.state_tag.to_string();
-    state_account.transition(clock, params, epoch_schedule)?;
-    if initial_state != state_account.state_tag.to_string() {
+    require!(
+        clock.epoch == steward_state.current_epoch,
+        StewardError::EpochMaintenanceNotComplete
+    );
```
ebatsell, line 45: Not sure if this reverted but the epoch check should be at the top of the state machine instructions https://github.com/jito-foundation/stakenet/pull/47#discussion_r1635544001

programs/steward/src/state/steward_state.rs
CoachChuckFF: nice catch - should be `num_pool_validators`

programs/steward/src/state/steward_state.rs
CoachChuckFF: I did this so we automatically ran the check everytime we ran a state machine function. Else, I'd have to put the check on each state machine function - which would be easy to forget to add the check

programs/steward/src/state/steward_state.rs
coachchucksol: Done

programs/steward/src/state/steward_state.rs
coachchucksol: Done

programs/steward/src/state/steward_state.rs
coachchucksol: Done

programs/steward/src/instructions/epoch_maintenance.rs
```diff
@@ -0,0 +1,95 @@
...
+
+    if (!state_account.state.checked_validators_removed_from_list).into() {
+        // Ensure there are no validators in the list that have not been removed, that should be
+        require!(
+            !check_validator_list_has_stake_status(
+                &ctx.accounts.validator_list,
+                StakeStatus::ReadyForRemoval
+            )?,
+            StewardError::ValidatorsHaveNotBeenRemoved
+        );
```
ebatsell, line 56: would be curious to see the CU cost on a MAX_VALIDATORS validator list, if it's not too big, might be worth just running this function for every IX to be able to get rid of the `checked_validators_removed` field

programs/steward/src/state/steward_state.rs
```diff
@@ -34,26 +34,29 @@ pub struct StateTransition {
...
         });
     }
     Ok(())
 }
 
 /// Tracks state of the stake pool.
-/// Follow state transitions here: [TODO add link to github diagram]
+/// Follow state transitions here:
+/// https://github.com/jito-foundation/stakenet/blob/master/programs/steward/state-machine-diagram.png
+// TODO Reorder fields to optimize for alignment
```
ebatsell, line 59: can remove this TODO line now

coachchucksol: Done

programs/steward/src/instructions/epoch_maintenance.rs
coachchucksol: Taking out for now - of need be we can add back in later

PR #48, opened 2024-06-12
Build IDL Anchor 0.30

ebatsell:
Fixes to build IDL for new Anchor 0.30 spec. New IDL spec breaking changes:
* `usize` not supported in instructions or accounts
* Need to implement IdlBuild for types that are non-native
  * Added manual deserialization functions for stake pool and validator list so we don't need to define them
* New features needed in build files


programs/steward/src/constants.rs
```diff
@@ -10,7 +10,7 @@ pub const EPOCH_PROGRESS_MAX: f64 = 0.99;
 // Cannot go more than 100 epochs without scoring
 pub const NUM_EPOCHS_BETWEEN_SCORING_MAX: u64 = 100;
 // Cannot score validators in under 100 slots, to submit 1 instruction per validator
-pub const COMPUTE_SCORE_SLOT_RANGE_MIN: usize = 100;
+pub const COMPUTE_SCORE_SLOT_RANGE_MIN: u64 = 100;
 #[cfg(feature = "mainnet-beta")]
 pub const VALIDATOR_HISTORY_FIRST_RELIABLE_EPOCH: usize = 520;
```
CoachChuckFF, line 15: Should we just change these right away to u64?

CoachChuckFF (commented):
What's the TODO here? Check the validator list address to the config?

CoachChuckFF (commented):
Maybe address the TODO

programs/steward/src/state/bitmask.rs
CoachChuckFF, line 1: Maybe make BITMASK_SIZE and all other const variables u64

programs/steward/src/state/steward_state.rs
CoachChuckFF, line 1: maybe make the index a uXX as well

programs/steward/src/constants.rs
CoachChuckFF, line 1: Make all constants uXX

programs/steward/src/state/bitmask.rs
ebatsell: For variables that are explicitly used for indexing into arrays or defining size I think we should keep them as usize, as long as they don't need to be exposed in the IDL

ebatsell:
> What's the TODO here? Check the validator list address to the config?

@CoachChuckFF I have a task in the "additional tweaks to steward" list to add validator_list to config, skipping for now since it may move around some data in the account so waiting until a future deploy of the program

programs/steward/src/state/steward_state.rs
```diff
@@ -388,9 +425,10 @@ impl StewardState {
             .num_pool_validators
             .checked_sub(1)
             .ok_or(StewardError::ArithmeticError)?;
+        let num_pool_validators = self.num_pool_validators as usize;
```
CoachChuckFF, line 428: This looks a lot nicer!

programs/steward/src/utils.rs
```diff
@@ -10,7 +10,7 @@ use spl_stake_pool::{
 
 use crate::{errors::StewardError, Config, Delegation};
 
-pub fn get_stake_pool(account: &AccountLoader<Config>) -> Result<Pubkey> {
+pub fn get_stake_pool_address(account: &AccountLoader<Config>) -> Result<Pubkey> {
```
CoachChuckFF, line 13: Good change

CoachChuckFF (approved):
All looks good to me!

coachchucksol (approved):
All looks good to me!

programs/steward/Cargo.toml
```diff
@@ -17,10 +17,10 @@ no-log-ix-name = []
 cpi = ["no-entrypoint"]
 default = ["custom-heap"]
 custom-heap = []
-# idl-build = ["anchor-lang/idl-build"]
+idl-build = ["anchor-lang/idl-build", "no-entrypoint"]
 
 [dependencies]
-anchor-lang = "0.30.0"
+anchor-lang = { features = ["idl-build"], version = "0.30.0" }
```
acheroncrypto, line 23: Hi, I was just passing by and saw you've enabled the `idl-build` feature for the default target. This should be avoided, as the IDL generation process is a separate build process that doesn't use Solana's build tools.

This seems harmless now other than adding a bit of unnecessary compilation, but it could very well break the compilation of your program in the future. For this reason, I added warnings for this usage in https://github.com/coral-xyz/anchor/pull/3061.

ebatsell: @acheroncrypto thanks for the input, will make the change!

PR #56, opened 2024-07-19
PATCH: Adds index mismatch interrupt

coachchucksol:
This PR will add in a service routine when a validator is removed. This functionality has been moved from epoch maintenance to its own routine to cover the case of validators being able to be removed on the same epoch if they are delinquent.


coachchucksol:
Problem:

When running the interrupt, there is currently no way of knowing on-chain which validator needs to be removed. Maybe we can pass in some account that says its been removed?

coachchucksol:
We can avoid adding in extra tracking indicies by looking at the validator list from smallest `marked_for_removal` to highest and check the corresponding state of the validator.

programs/steward/src/instructions/idle.rs
```diff
@@ -14,6 +17,16 @@ pub struct Idle<'info> {
...
+
+    /// CHECK: Correct account guaranteed if address is correct
+    #[account(address = deserialize_stake_pool(&stake_pool)?.validator_list)]
+    pub validator_list: AccountInfo<'info>,
+
+    /// CHECK: Correct account guaranteed if address is correct
+    #[account(
+        address = get_stake_pool_address(&config)?
+    )]
+    pub stake_pool: AccountInfo<'info>,
```
ebatsell, line 29: we added the validator list address to the config account so this isn't necessary

ebatsell: a few other instructions where we can remove stake_pool as well

PR #60, opened 2024-07-24
Fix gossip service connection for Validator Keeper

ebatsell:
Fixes the way that we connect to gossip service by using the method from solana-gossip to start the gossip service and passing in another IP.
Still unclear why it suddenly stopped working.


ebatsell:
merged in when christian/keeper-branch was merged

PR #62, opened 2024-07-24
TWEAKS: Last cleanup

coachchucksol:


programs/steward/src/instructions/rebalance.rs
```diff
@@ -163,6 +162,21 @@ pub fn handler(ctx: Context<Rebalance>, validator_list_index: usize) -> Result<(
         validator_stake_info.vote_account_address == validator_history.vote_account,
         StewardError::ValidatorNotInList
     );
+
+    // Checking Vote Account
+    require!(
+        validator_stake_info.vote_account_address == ctx.accounts.vote_account.key(),
+        StewardError::VoteAccountDoesNotMatch
```
ebatsell, line 169: Could arguably delete the check above if we have this because the validator history address has seeds checking with the vote account

programs/steward/src/lib.rs
```diff
@@ -189,19 +193,20 @@ pub mod steward {
         instructions::close_steward_accounts::handler(ctx)
     }
 
-    /* Passthrough instructions to spl-stake-pool, where the signer is Staker. Must be invoked by `config.authority` */
```
ebatsell, line 192: I actually like having the visual separation between the above and the passthrough instructions

programs/steward/src/score.rs
```diff
@@ -206,7 +205,7 @@ pub fn validator_score(
 
     /////// Formula ///////
 
-    let yield_score = (average_vote_credits / average_blocks) * (1. - commission);
+    let yield_score = (average_vote_credits / average_blocks.max(1.0)) * (1. - commission);
```
ebatsell, line 208: thought we'd let this fail here?

programs/steward/src/instructions/rebalance.rs
coachchucksol: Good point

programs/steward/src/score.rs
coachchucksol: you're right

programs/steward/src/lib.rs
coachchucksol: readded

PR #72, opened 2024-08-05
Fixes: rebalance decrease amount calculation, validator_lamport_balance initialization, tweaks to keeper

ebatsell:
Changes:
* Fixes rebalance decrease: the stake pool's validator_list lamport amount can get out of alignment with the associated stake account's active lamports, resulting in more lamports being withdrawn from validators than is possible. The problem is fixed by passing along the true active stake amount into the decrease_validator_stake calculation
* Unblocks unstaking for stake deposits by fixing how validator_lamport_balances are initialized. Currently, all validator_lamport_balances start at 0, meaning all current lamports are seen as stake deposits and tried to be unstaked until the cap is hit. However, we can load each value in a `rebalance` instruction when there are no activating or deactivating lamports and then continue with the working process. Adds a sentinel value and checks to skip modifying it when sentinel is present.
* Will deploy on test program sssh4zkKhX8jXTNQz1xDHyGpygzgu2UhcRcUvZihBjP before deploying on mainnet program


programs/steward/src/state/steward_state.rs
```diff
@@ -967,15 +967,6 @@ impl StewardState {
                 RebalanceType::None
             };
 
-            msg!("Reserve lamports after adjustment: {}", reserve_lamports);
```
CoachChuckFF, line 970: Too much CU?

programs/steward/src/state/steward_state.rs
```diff
@@ -1026,9 +1017,12 @@ impl StewardState {
                     }
                 }
                 RebalanceType::Increase(amount) => {
-                    self.validator_lamport_balances[index] = self.validator_lamport_balances[index]
-                        .checked_add(amount)
-                        .ok_or(StewardError::ArithmeticError)?;
+                    if self.validator_lamport_balances[index] != LAMPORT_BALANCE_DEFAULT {
```
CoachChuckFF, line 1020: Do we want this same check for Decrease?

CoachChuckFF (changes requested):
A couple of questions needed for clarification

programs/steward/src/state/steward_state.rs
ebatsell: was causing local tests to fail since it was calling msg! outside of context, + think we're okay to remove these now since we have the emits

programs/steward/src/delegation.rs
```diff
@@ -245,6 +252,7 @@ impl UnstakeState {
         // either to the target or to the previous balance before the deposit, whichever is lower in terms of total lamports unstaked
         if current_lamports > state.validator_lamport_balances[index]
             && self.stake_deposit_unstake_total < self.stake_deposit_unstake_cap
+            && state.validator_lamport_balances[index] != LAMPORT_BALANCE_DEFAULT
```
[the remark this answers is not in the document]
ebatsell: the increase_stake_calculation algo here doesn't interact with this field so we're safe

programs/steward/src/state/steward_state.rs
ebatsell: yep, good catch

programs/steward/src/delegation.rs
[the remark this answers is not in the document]
CoachChuckFF: 10-4

programs/steward/src/state/steward_state.rs
CoachChuckFF: 10-4

CoachChuckFF (approved):
lgtm

PR #76, opened 2024-08-12
Finalize keeper, improve tx confirmation

ebatsell:
* use \_with\_history for confirmation otherwise txs expire after 300 blocks


coachchucksol (approved):
All looks great! Glad to see you fixed the test script!

PR #81, opened 2024-09-10
Emit scoring details and instant unstake details in V2 events

ebatsell:
Updates scoring and instant unstake functions with several improvements:
* Events now contain a "details" field which, for each given component that fails or is triggered, will now track the specific epoch and value that contained the violation, and will be emitted.
   * Will be shown in UI to improve validators' ability to diagnose scoring failures
   * Both response structs are now updated to a V2 since adding new fields without changing the name would break the ability to backfill the event from genesis, as the event discriminator is based on the name, and fields changing mid-history is a very annoying thing to code in. Old structs are stored and noted
 * Each individual component and the associated values are now calculated in their own helper methods which greatly improves readability.
   * In a future PR we can add tests for these helper functions, which will allow us to really simplify the unit tests for these functions in`test_algorithms.rs`
* All score functionality remains the same as confirmed by tests


programs/steward/src/score.rs
```diff
@@ -47,194 +49,348 @@ pub struct ScoreComponents {
...
+fn calculate_mev_commission(
+    mev_commission_window: &[Option<u16>],
+    current_epoch: u16,
+    mev_commission_bps_threshold: u16,
+) -> (f64, u16, u16, f64) {
+    let (max_mev_commission, max_mev_commission_epoch) = mev_commission_window
+        .iter()
+        .rev()
+        .enumerate()
+        .filter_map(|(i, &commission)| commission.map(|c| (c, current_epoch - i as u16)))
```
coachchucksol, line 208: Might want to make this a checked_sub

coachchucksol (approved):
lgtm

PR #82, opened 2024-09-10
Force state transitions to Idle before doing work in new epoch

ebatsell:
Currently a single lingering ComputeInstantUnstake or Rebalance step can be executed if the Steward was left in this state after rolling over to a new epoch. This can affect stake based on outdated information which we don't want. Force the transition to Idle so everything is handled normally.


coachchucksol (approved):
lgtm

PR #83, opened 2024-09-10
Blacklist instructions accept multiple validators

ebatsell:
Update both adding and removing instructions to take in a Vec<u32> of validator history indices for updating the blacklist. This makes multisig IXs for bulk adding to the blacklist much easier.


coachchucksol (changes requested):
Add a test or two adding and removing blacklists

coachchucksol (approved):
lgtm

PR #84, opened 2024-09-13
Fix for removed validator past boundary

ebatsell:
Problem: A bug was found where validators who are added to the pool outside of the boundary of the Steward state, then removed before they begin getting tracked (validators added during cycle N are tracked in steward state in cycle N+1) were not getting properly removed. This caused the state machine to get stuck because the invariant used to keep the state aligned was failing (`num_pool_validators + validators_added - validators_to_remove == validator_list.len()`)

Resolution: Make sure we're shifting values for steward state fields that can be relevant past `num_pool_validators` (which is just the validators_to_remove and validators_for_immediate_removal). We also don't want to clear those values at `num_pool_validators` because that might be a relevant value.

### DIFF
--- a/keepers/stakenet-keeper/src/entries/crank_steward.rs
+++ b/keepers/stakenet-keeper/src/entries/crank_steward.rs
@@ -266,7 +266,7 @@ async fn _handle_instant_removal_validators(
 
     while validators_to_remove.count() != 0 {
         let mut validator_index_to_remove = None;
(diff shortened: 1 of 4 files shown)


coachchucksol (approved):
lgtm - My only note is it took me a bit to convince myself that setting all values at `index_to_clear` is valid. IE no value set outside of num_pool_validators is being used, and will be reset on the next cycle.

PR #86, opened 2024-09-17
No idl-build in default target

ebatsell:
Recommendation from #48. Should help compilation issues downstream.


coachchucksol (approved):
No issues on my end

PR #92, opened 2024-12-01
Normalize Timely Vote Credits in Steward score

ebatsell:
Normalizes vote credits which are now 8x higher for validators starting in epoch 704.

### DIFF
--- a/programs/steward/src/score.rs
+++ b/programs/steward/src/score.rs
@@ -274,7 +274,7 @@ pub fn calculate_epoch_credits(
             // If vote credits are None, then validator was not active because we retroactively fill credits for last 64 epochs.
             // If total blocks are None, then keepers missed an upload and validator should not be punished.
             let credits = maybe_credits.unwrap_or(0);
-            let ratio = credits as f64 / *blocks as f64;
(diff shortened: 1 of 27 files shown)


programs/validator-history/src/state.rs
```diff
@@ -274,10 +275,49 @@ impl CircBuf {
         field_latest!(self, epoch_credits)
     }
 
+    /// Normalized epoch credits, accounting for Timely Vote Credits making the max number of credits 8x higher
+    /// for every epoch starting at `tvc_activation_epoch`
```
ebatsell, line 279: add a comment listing activation epochs

programs/steward/src/score.rs
```diff
@@ -90,6 +90,7 @@ pub fn validator_score(
     cluster: &ClusterHistory,
     config: &Config,
     current_epoch: u16,
```
coachchucksol, line 92: current_epoch should be u64

coachchucksol (approved):
lgtm

programs/steward/src/score.rs
ebatsell: legacy issue cause of ValidatorHistoryEntry trying to minimize bytes

PR #94, opened 2025-02-19
feat: Add tip router tracking and scoring

tomjohn1028:
* `ValidatorHistoryEntry` now tracks the _merkle_root_upload_authority_
* Adds _tip_router_upload_auth_epoch_cutoff_ flag to `Config` which designated when only TipRouter uploads are acceptable
* Steward includes a binary score related to merkle root upload authority _merkle_root_upload_authority_score_
* Steward will allow instant unstake if there's an unacceptable _merkle_root_upload_authority_


programs/steward/src/score.rs
```diff
@@ -611,3 +645,29 @@ pub fn calculate_instant_unstake_blacklist(config: &Config, validator_index: u32
...
+                    Ok(true)
+                } else {
+                    Ok(false)
+                }
+            }
+            MerkleRootUploadAuthority::TipRouter => Ok(false),
+            _ => Ok(true),
+        }
+    } else {
+        // Default to 0 if empty history?
```
tomjohn1028, line 670: @ebatsell check if this is correct. Sets instant unstake to true, and thus score to 0, if there's no history

programs/steward/src/state/accounts.rs
```diff
@@ -44,8 +44,15 @@ pub struct Config {
     /// Halts any state machine progress
     pub paused: U8Bool,
 
+    /// Required for 8 byte alignment
```
ebatsell, line 47: could we try PodU16 here actually to avoid the split up padding?

tests/tests/steward/test_algorithms.rs
```diff
@@ -696,6 +714,187 @@ fn test_compute_score() {
...
+                delinquency_epoch: EPOCH_DEFAULT,
+                max_commission: 0,
+                max_commission_epoch: 10,
+                max_historical_commission: 0,
+                max_historical_commission_epoch: 0,
+            }
+        }
+    );
+
+    // test merkle_root_upload_authority is OldJito at tip_router_upload_auth_epoch_cutoff
```
ebatsell, line 763: nit: before epoch_cutoff
also rm println below

programs/steward/src/score.rs
```diff
@@ -422,9 +433,22 @@ pub fn calculate_blacklist(config: &Config, validator_index: u32) -> Result<f64>
     }
 }
 
+/// Checks if validator is using appropriate TDA MerkleRootUploadAuthority
+pub fn calculate_merkle_root_authoirty(
```
ebatsell, line 437: nit: authoirty

programs/steward/src/score.rs
ebatsell: let's set to false to be conservative. there are be plenty of other mechanisms that prevent a validator with no history from getting stake, we don't want this to be the hidden linchpin

ebatsell (commented):
A few nits, looks great

programs/steward/src/state/accounts.rs
tomjohn1028: Not able to drop in with PodU16. Any idea why [`PodU16`](https://docs.rs/spl-pod/0.5.0/src/spl_pod/primitives.rs.html#70) doesn't implement BorshSerialize like [`PodU32`](https://docs.rs/spl-pod/0.5.0/src/spl_pod/primitives.rs.html#87) does? I ended up manually implementing the BorshSerialize trait. 


https://github.com/jito-foundation/stakenet/pull/94/commits/e7ff4ef6c2a9ee5417aea3347ba64d1b1cf3b91b

tomjohn1028: updated https://github.com/jito-foundation/stakenet/pull/94/commits/f3c7c83b2ef16de7e232c911eac1268b5aa66ea8

programs/validator-history/src/state.rs
```diff
@@ -64,7 +135,9 @@ pub struct ValidatorHistoryEntry {
     pub vote_account_last_update_slot: u64,
     // MEV earned, stored as 1/100th SOL. mev_earned = 100 means 1.00 SOL earned
     pub mev_earned: u32,
-    pub padding1: [u8; 84],
+    /// The enum mapping of the Validator's Tip Distribution Account's merkle root upload authority
+    pub merkle_root_upload_authority: MerkleRootUploadAuthority,
```
tomjohn1028, line 139: Should we actually updated `padding0` to be the `MerkleRootUploadAuthority` since it's a u8?

programs/steward/src/state/accounts.rs
```diff
@@ -44,8 +48,12 @@ pub struct Config {
     /// Halts any state machine progress
     pub paused: U8Bool,
 
+    /// The epoch after which validators must be using TipRouter upload authority for tip
+    /// distribution
+    pub tip_router_upload_auth_epoch_cutoff: PodU16,
```
ebatsell, line 53: We should actually get rid of this, extra risk for very little added benefit. Risk of people getting very upset if they don't update in time even though they're still effectively using tip router

ebatsell (commented):
Realized we should remove the epoch cutoff check.

After that will be ready to put into a feature branch. Can keep it there until we're ready to release so it'll be easy to diff all changes with master

tomjohn1028:
Closing to use feature branch `feat/add-priority-fee-support`

PR #98, opened 2025-04-14
Priority Fee Oracle keeper

tomjohn1028:


keepers/stakenet-keeper/src/operations/block_metadata/operations.rs
```diff
@@ -0,0 +1,428 @@
...
+    epoch: u64,
+    starting_slot: u64,
+    ending_slot: u64,
+) {
+    debug!(
+        "Gathering data for slots: {} - {}",
+        starting_slot, ending_slot
+    );
+    let rpc_leader_schedule = rpc_client
+        .get_leader_schedule(Some(starting_slot))
```
ebatsell, line 217: Curious if this is backed by bigtable storage in the rpc, otherwise what happens if we fetch a leader schedule from several epochs ago

keepers/stakenet-keeper/src/operations/block_metadata/operations.rs
```diff
@@ -0,0 +1,428 @@
...
+        None => {
+            let mut leader_info = AggregateBlockInfo::new(epoch);
+            leader_info.increment_data(leader_slots, blocks_produced, priority_fees);
+            leaders_map.insert(leader.clone(), leader_info);
+        }
+    }
+}
+
+/// Iterates through each slot from _starting_slot_ to _ending_slot_, inclusive of the ending slot.
+/// Lots the block metadata and aggreates the information on a per leader basis.
```
ebatsell, line 345: Noting this just works for start-end slots within a single epoch because it's just keyed by leader, maybe leader-epoch key would be more flexible

keepers/stakenet-keeper/src/operations/block_metadata/operations.rs
```diff
@@ -0,0 +1,428 @@
...
+                // get the priority fee rewards for the block.
+                let maybe_block_rewards = block
+                    .rewards
+                    .unwrap()
+                    .into_iter()
+                    .find(|r| r.reward_type == Some(RewardType::Fee));
+                let priority_fees = match maybe_block_rewards {
+                    Some(block_reward) => block_reward.lamports,
+                    None => 0,
+                };
```
ebatsell, line 371: Just confirming logic is equivalent to Christian's implementation: https://github.com/jito-foundation/jito-solana/blob/ck/distro-script/priority-fee-sharing/src/lib.rs#L127

ebatsell: Worth double checking we never have multiple entries in the iter with reward_type == RewardType::Fee

keepers/stakenet-keeper/src/operations/block_metadata/operations.rs
```diff
@@ -0,0 +1,428 @@
...
+            RpcBlockConfig {
+                encoding: Some(UiTransactionEncoding::Json),
+                transaction_details: Some(TransactionDetails::None),
+                rewards: Some(true),
+                commitment: Some(CommitmentConfig::finalized()),
+                max_supported_transaction_version: Some(0),
+            },
+        )
+        .await;
+    let block = match block_res {
```
ebatsell, line 403: * Add fallback rpc calls to confirm block was indeed missed by network
* Add comment here outlining the cases that might result in "SkippedBlock"

Just remembered: another useful thing to confirm missing block is the SlotHistory sysvar, which stores latest 1,048,576 slots in a BitMap so you can determine if a slot was actually missing. This is the best source of truth for slots < 2 epochs old

ebatsell (approved):

keepers/stakenet-keeper/src/operations/block_metadata/operations.rs
tomjohn1028: So `aggregate_information` can handle more than 1 epoch?

Since the SQLite record's are on a per epoch basis (just like each ValidatorHistoryEntry), it feels cleaner to handle each epoch as it's own chunk and multi-epoch handling can be done higher up in the control flow.

keepers/stakenet-keeper/src/operations/block_metadata/operations.rs
tomjohn1028: I spot checked blocks and never saw it. Would it be better to sum?

keepers/stakenet-keeper/src/operations/block_metadata/operations.rs
tomjohn1028: > SlotHistory sysvar
genius

keepers/stakenet-keeper/src/operations/block_metadata/operations.rs
tomjohn1028: RPC method only checks an in memory cache. But it looks like that cache holds 10 epochs worth of schedules by default https://github.com/anza-xyz/agave/blob/v2.1.20/ledger/src/leader_schedule_cache.rs#L21-L39

keepers/stakenet-keeper/src/state/keeper_state.rs
```diff
@@ -117,6 +122,20 @@ pub struct KeeperState {
     pub steward_progress_flags: StewardProgressFlags,
 }
 impl KeeperState {
+    // REVIEW: We could put this into the get_vote_account_map function and reduce the iteration
```
ebatsell, line 125: Current approach seems fine

keepers/stakenet-keeper/src/state/update_state.rs
```diff
@@ -53,13 +55,36 @@ pub async fn pre_create_update(
...
+
     // Fetch Vote Accounts
     keeper_state.vote_account_map = get_vote_account_map(client).await?;
 
     // Get all get vote accounts
     keeper_state.all_get_vote_account_map =
         get_all_get_vote_account_map(client, keeper_state).await?;
 
+    // Update mapping of identity keypair to vote account
+    keeper_state.update_identity_to_vote_map();
```
ebatsell, line 86: super nit: move this to right after `keeper_state.vote_account_map = ` line

keepers/stakenet-keeper/src/operations/block_metadata/operations.rs
```diff
@@ -0,0 +1,565 @@
...
+                                }
+                                slot_history::Check::TooOld | slot_history::Check::Found => {
+                                    // REVIEW: Should we handle TooOld and Found differently?
+                                    if let Some(redundant_rpc_urls) = maybe_redundant_rpc_urls {
+                                        if redundant_rpc_index >= redundant_rpc_urls.len() {
+                                            return Err(BlockMetadataKeeperError::SkippedBlock);
+                                        }
+                                        current_client = &redundant_rpc_urls[redundant_rpc_index];
+                                        redundant_rpc_index += 1;
+                                        continue;
```
ebatsell, line 544: does the redundant rpc urls actually do anything here? current_client and redundant_rpc_index are just local to this function - was thinking we would iterate through these to fetch blocks if 1) slot history says a block exists but original RPC didn't find it or 2) slot history is too old so we need to rely on these to figure out if block existed

tomjohn1028: I believe this achieves the goal

1. If current_client thinks slot was skipped, check `SlotHistory` for the slot. ([source](https://github.com/jito-foundation/stakenet/blob/tj/keeper-block-data/keepers/stakenet-keeper/src/operations/block_metadata/operations.rs#L527-L529))
2. If the `SlotHistory` check returns TooOld or Found && _redundant_rpc_urls_ exist && we have not reached the end of _redundant_rpc_urls_, lets set the current_client to the next RPC client, increment the index, and continue the loop. where current_client calls get_block. ([source](https://github.com/jito-foundation/stakenet/blob/tj/keeper-block-data/keepers/stakenet-keeper/src/operations/block_metadata/operations.rs#L536-L538))

If at any point a block is returned by the RPC or a different error is produced, we short circuit returning the block or the error.

ebatsell: Makes sense now, I missed the original loop

ebatsell (commented):
lgtm

PR #99, opened 2025-04-15
validator-history-cli: view-config and update-oracle-authority commands

gzalz:
**Problem**
The validator history CLI lacks the ability to display the program config as well as the ability to change the oracle authority with a permissioned call from the admin.

**Solution**
- Add `update-oracle-authority`: Permissioned instruction to update the oracle authority, giving the provided pubkey permissions to modify protected fields of ValidatorHistory accounts
- Add `view-config`: Display various accounts and state of the validator history program


ebatsell (dismissed):

.github/workflows/build.yaml
```diff
@@ -164,7 +164,7 @@ jobs:
   # after waiting, it will download the verified build and IDL from the verified_build step and release the artifacts
   release:
     name: release
-    runs-on: big-runner-1
+    runs-on: ubuntu-latest
```
ebatsell, line 167: Should revert the build changes

mrmizz (approved):

PR #106, opened 2025-04-27
feat: add priority fee commission score to steward

tomjohn1028:


programs/steward/src/score.rs
```diff
@@ -9,15 +9,16 @@ use validator_history::{
...
+        VALIDATOR_HISTORY_FIRST_RELIABLE_EPOCH,
     },
     errors::StewardError::{self, ArithmeticError},
     Config,
 };
 
 #[event]
 #[derive(Debug, PartialEq)]
-pub struct ScoreComponentsV3 {
+pub struct ScoreComponentsV4 {
```
ebatsell, line 21: We actually haven't deployed the upgrade from V2-> V3 yet so can keep this as V3

programs/steward/src/state/parameters.rs
```diff
@@ -172,9 +201,20 @@ pub struct Parameters {
...
+    /// (current_epoch - offset - pf_lookback_epochs) to (current_epoch - offset)
+    pub pf_lookback_offset: u8,
+
+    /// The maximum validator commission before the validator scores 0.
+    /// E.g. 5_000 bps (50%) would mean: if the validator keeps > 50% of priority fees,
+    /// then score = 0
+    pub pf_max_commission_bps: u16,
+
+    /// An error of margin for priority fee commission calculations
+    pub pf_error_margin_bps: u16,
```
ebatsell, line 217: Clean 6 bytes

programs/steward/src/state/accounts.rs
```diff
@@ -59,6 +68,15 @@ impl Config {
...
         self.paused = paused.into();
     }
+
+    /// Determine the realized tip threshold for a validator given it's `total_priority_fees`
+    pub fn priority_fee_tip_threshold(&self, total_priority_fees: u64) -> u64 {
+        ((10_000 - u64::from(self.parameters.pf_max_commission_bps)
+            + u64::from(self.parameters.pf_error_margin_bps))
+            * total_priority_fees
+            + 9_999)
+            / 10_000
```
ebatsell, line 78: should use BASIS_POINT_MAX here

also, what's the purpose of the 9999? to round up?

programs/steward/src/score.rs
```diff
@@ -440,6 +461,90 @@ pub fn calculate_merkle_root_authority(validator: &ValidatorHistory) -> Result<f
...
+                        start_epoch + u16::try_from(relative_epoch).unwrap();
+                    (commission_bps, max_commission_epoch)
+                } else {
+                    agg
+                }
+            },
+        );
+
+    // return score 1 when there's not enough history
+    if priority_fee_tips.first().is_none() || total_priority_fees.first().is_none() {
```
ebatsell, line 511: If priority fee tips.first() is none but total_priorirty fees isn't, pretty sure that implies there was no fees shared because there was no PFDA, so we should only include total_priority_fees.first().is_none() in this condition

programs/steward/src/score.rs
```diff
@@ -440,6 +461,90 @@ pub fn calculate_merkle_root_authority(validator: &ValidatorHistory) -> Result<f
...
+        return Ok((
+            1.0,
+            max_priority_fee_commission,
+            max_priority_fee_commission_epoch,
+        ));
+    }
+
+    // REVIEW: This is handled through an aggregated commission. This leads to valdiators with a
+    // drastic change in stake and commission gets smoothed out. Should this scoring be done
+    // normalized by actual commission bps instead?
```
ebatsell, line 521: Good point, if a validator accidentally under-shared for an epoch and then lost a lot of stake, there would be no way to over-pay and get back into good graces

I think calculating the commission rate per epoch and then averaging those values would be better 

Curious for your thoughts @coachchucksol 

Curious your thoughts @coachchucksol

programs/steward/src/state/accounts.rs
tomjohn1028: > also, what's the purpose of the 9999? to round up?
Correct, it's a rounding technique. Figure in this case the threshold needs to go in favor of the staker

programs/steward/src/score.rs
tomjohn1028: Digging in here a bit more and correct me if I'm wrong:

ValidatorHistoryEntries use default values of MAX. So _None_ will only occur if the entire `ValidatorHistoryEntry` does not exist. So I think the case where one of these is none, while the other is some, shouldn't be possible.

Now the fact that we set _priority_fee_tips_ in `CopyPriorityFeeDistribution` and _total_priority_fees_ in `UpdatePriorityFeeHistory`, means there is a chance where _ priority_fee_tips_ is MAX and _total_priority_fees_ is some value. I think this is the state that would mean there is no PFDA, and it must be handled.

tomjohn1028: Correction, I see that the macros like _field_range_ has the following code and converts the MAX value to None `.filter(|&field| field != ValidatorHistoryEntry::default().$field)`

programs/steward/src/score.rs
```diff
@@ -440,6 +461,97 @@ pub fn calculate_merkle_root_authority(validator: &ValidatorHistory) -> Result<f
...
+        .checked_mul(BASIS_POINTS_MAX_U64)
+        .and_then(|scaled_tips| scaled_tips.checked_div(total_fees))
+        .unwrap_or(0);
+    BASIS_POINTS_MAX
+        .checked_sub(u16::try_from(inverse_commission_bps).unwrap_or(0))
+        .unwrap_or(BASIS_POINTS_MAX)
+}
+
+/// Checks if validator is maintaining < X% realized commission rates over some history of epochs
+pub fn calculate_priority_fee_commission(
```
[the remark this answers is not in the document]
coachchucksol: I spoke too early. You have a testing file :)

coachchucksol (commented):
Looking good! Mostly nits

coachchucksol:
@tomjohn1028 a couple guidlines:

- Don't use `unwrap`
- Don't use abbreviations - keep names verbose
- Use `check_xxx` for all arithmetic
- Keep things as single responsibility as possible

programs/steward/src/score.rs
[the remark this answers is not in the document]
tomjohn1028: Gonna resolve this one. Feel free to re-open if you find additional tests are needed.

programs/steward/src/score.rs
coachchucksol: This sounds good to me

ebatsell (approved):
LGTM

PR #117, opened 2025-05-26
Fix RUSTSEC-2021-0141

aoikurokawa:
- `dotenv` is not maintained anymore, so we use `dotenvy` instead


aoikurokawa:
Thank you for reviewing @gzalz

PR #119, opened 2025-05-28
fix: pin cargo-sort to compatible version

gzalz:
**Problem**
CI fails because cargo-sort is not pinned to a version and the latest release of cargo-sort uses features that are unstable in our CI's version of rust.

**Solution**
- Pin cargo-sort to 1.0.9


gzalz:
Addressed in @aoikurokawa 's diff

PR #126, opened 2025-07-02
feat: validator history priority fee support

gzalz:
```
test steward::test_cycle::test_cycle ... ok

test result: ok. 126 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 2.34s

   Doc-tests tests

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s
```

### DIFF
--- a/programs/validator-history/src/state.rs
+++ b/programs/validator-history/src/state.rs
@@ -75,10 +75,11 @@ static_assertions::const_assert_eq!(size_of::<ValidatorHistoryEntry>(), 128);
 #[repr(u8)]
 pub enum MerkleRootUploadAuthority {
     #[default]
-    Empty,
-    Other,
-    OldJitoLabs,
-    TipRouter,
+    Unset = 0,
+    Empty = u8::MAX,
+    Other = 1,
+    OldJitoLabs = 2,
+    TipRouter = 3,
 }
 
 unsafe impl Zeroable for MerkleRootUploadAuthority {}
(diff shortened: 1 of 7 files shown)


programs/validator-history/src/state.rs
```diff
@@ -191,20 +200,22 @@ impl Default for ValidatorHistoryEntry {
             rank: u32::MAX,
             vote_account_last_update_slot: u64::MAX,
             mev_earned: u32::MAX,
-            merkle_root_upload_authority: MerkleRootUploadAuthority::default(),
+            merkle_root_upload_authority: MerkleRootUploadAuthority::Unset,
```
ebatsell, line 203: We either want Unset to be u8::MAX or for default to be Empty and remove the Unset variant. Since the accounts are pre-loaded with all 1s, this would be incorrect

programs/validator-history/src/state.rs
```diff
@@ -103,6 +103,10 @@ impl IdlBuild for MerkleRootUploadAuthority {
             name: "MerkleRootUploadAuthority".to_string(),
             ty: IdlTypeDefTy::Enum {
                 variants: vec![
+                    IdlEnumVariant {
+                        name: "Unset".to_string(),
+                        fields: None,
+                    },
```
ebatsell, line 109: Nit: "Empty" variant still exists in IDL

ebatsell (approved):

PR #134, opened 2025-07-30
fix: Priority fee copy and scoring logic

gzalz:
**Problem**
Current copy_priority_fee_distribution_account and compute_score instruction handlers don't consider all edge cases considering their permissionless nature.

- There is no way to distinguish a PFDA that has not been copied vs. a PFDA that does not exist (no PF sharing opt-in from validator)
- PFDAs can be copied before they are finalized
- Scoring can be invoked before we can guarantee a PFDA has been copied potentially causing erroneous unstake conditions for a validator

**Solution**
- Disallow calling copy_priority_fee_distribution_account until the epoch that account tracks is over
- Allow copying PFDAs that do not exist - this is the only way we can distinguish DNE from not yet copied on-chain
- Add new DNE merkle_root_upload_authority enum value
- Disallow re-copying a specific PFDA to remove any possibility of copying a reclaimed account

**IMPORTANT**
This diff has been modified to have priority fee related score defaulted to 1.0. The intended logic has been commented out until the time of priority fee scoring launch.


programs/validator-history/src/state.rs
```diff
@@ -749,7 +753,7 @@ impl ValidatorHistory {
             epoch,
             priority_fee_commission: commission,
             priority_fee_tips,
-            merkle_root_upload_authority,
```
gzalz, line 752: This was wrong.

ebatsell: great catch

programs/validator-history/src/instructions/copy_priority_fee_distribution.rs
```diff
@@ -41,7 +41,6 @@ pub struct CopyPriorityFeeDistribution<'info> {
         ],
         bump,
         seeds::program = config.priority_fee_distribution_program.key(),
-        owner = config.priority_fee_distribution_program.key()
```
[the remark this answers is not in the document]
mrmizz: Oh even better. Did not quite catch that. Nice work.

programs/validator-history/src/state.rs
ebatsell: for the record: this bug existed for the case where the copy_priority_fee_distribution ix is called on a validator as the first instruction before any other validator history ixs are called and there exists no entry for this instruction

programs/validator-history/src/instructions/copy_priority_fee_distribution.rs
```diff
@@ -57,14 +56,38 @@ pub fn handle_copy_priority_fee_distribution_account(
     if epoch > Clock::get()?.epoch {
         return Err(ValidatorHistoryError::EpochOutOfRange.into());
     }
+
+    // The PFDA cannot be updated outside of its own epoch - this guarantees the immutability of the data we copy
+    if epoch == Clock::get()?.epoch {
+        return Err(ValidatorHistoryError::PriorityFeeDistributionAccountNotFinalized.into());
+    }
```
ebatsell, line 63: I don't understand why we want this? Fees can be transfered to the PFDA throughout its epoch, it can be created up until the last slot as well. But in the next epoch, everything is guaranteed to be immutable

programs/validator-history/src/instructions/copy_priority_fee_distribution.rs
```diff
@@ -57,14 +56,38 @@ pub fn handle_copy_priority_fee_distribution_account(
...
+            return Err(ValidatorHistoryError::PriorityFeeDistributionAccountAlreadyCopied.into());
+        }
+    }
+
     let mut pdfa_data: &[u8] = &ctx.accounts.distribution_account.try_borrow_data()?;
 
-    let distribution_account = PriorityFeeDistributionAccount::try_deserialize(&mut pdfa_data)?;
+    let distribution_account =
+        PriorityFeeDistributionAccount::try_deserialize(&mut pdfa_data).unwrap_or_default();
+    // If the distribution account is not found, we set the default values of 0 for the commission and priority fees earned
```
ebatsell, line 86:
```suggestion
    // If the distribution account is not found, we set the default values of 0 for the commission and priority fees transferred
```

programs/validator-history/src/instructions/copy_priority_fee_distribution.rs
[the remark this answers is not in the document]
ebatsell: Was copying a non-existent account possible before? I think it wouldn't have been because of the owner constraint

programs/validator-history/src/instructions/copy_priority_fee_distribution.rs
ebatsell: and the function name

programs/validator-history/src/state.rs
```diff
@@ -1164,6 +1168,15 @@ impl ValidatorHistory {
         self.history.push(entry);
         Ok(())
     }
+
+    /// Boolean representing if fields required for scoring have been set
+    pub fn can_score(&self) -> bool {
+        !matches!(
+            self.history
+                .priority_fee_merkle_root_upload_authority_latest(),
+            Some(MerkleRootUploadAuthority::Unset)
```
ebatsell, line 1177: The timing here doesn't make sense? `_latest()` means "current epoch's entry", which, if we have the requirement that a PFDA can't be copied in its epoch, is conflicting

Also the approach needs to be more thorough: we shouldn't just have this can_score requirement for scoring. We want to update the way priority fee commission is calculated by checking within that algo to determine if *each* epoch's PFDA is missing, or if the PFDA was simply not copied by validator history keeper. If pfda was not copied, skip that epoch. If missing, shared = 0

ebatsell (changes requested):

programs/validator-history/src/state.rs
gzalz: Removing to move logic into the pf commission calculation.

programs/validator-history/src/instructions/copy_priority_fee_distribution.rs
```diff
@@ -57,14 +56,38 @@ pub fn handle_copy_priority_fee_distribution_account(
     if epoch > Clock::get()?.epoch {
         return Err(ValidatorHistoryError::EpochOutOfRange.into());
     }
+
+    // The PFDA cannot be updated outside of its own epoch - this guarantees the immutability of the data we copy
```
coachchucksol, line 60: nit comment: The PDFA's priority_fee data is not valid until epoch n+1..

programs/validator-history/src/state.rs
```diff
@@ -1164,6 +1168,15 @@ impl ValidatorHistory {
         self.history.push(entry);
         Ok(())
     }
+
+    /// Boolean representing if fields required for scoring have been set
+    pub fn can_score(&self) -> bool {
```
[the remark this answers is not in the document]
coachchucksol: I jive with this ^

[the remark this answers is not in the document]
coachchucksol: I would actually name it something a little more verbose, `has_copied_priority_fees()`

In this case `can_score()` seems to vague, and `is_finalized()` is closer, but its meaning is lost.

tests/tests/validator_history/test_copy_priority_fee_distribution.rs
```diff
@@ -122,23 +173,49 @@ async fn test_priority_fee_commission_fail() {
...
+    // assert values, Priority Fee earned default
+    let account: ValidatorHistory = fixture
+        .load_and_deserialize(&fixture.validator_history_account)
         .await;
+    assert!(account.history.idx == 0);
+    assert!(account.history.arr[0].epoch == 0);
+    assert!(account.history.arr[0].priority_fee_commission == 42);
+    assert!(account.history.arr[0].priority_fee_tips == priority_fees_earned);
+    assert!(
+        account.history.arr[0].priority_fee_merkle_root_upload_authority
```
coachchucksol, line 197: Can you also makes a test that's only function is to make sure that all of the MerkleRootUploadAuthority enums are possible?

IE, have a history account that matches the 4 enums.

coachchucksol (changes requested):
This all seems to hit what we need!

- Cannot call copy on N epoch
- Cannot overwrite data once its been written
- Added the DNE
- Cannot score without a valid MRUA

Just a couple of comments

programs/validator-history/src/instructions/copy_priority_fee_distribution.rs
```diff
@@ -57,20 +56,44 @@ pub fn handle_copy_priority_fee_distribution_account(
...
+    if let Some(entry) = validator_history_entry_for_epoch {
+        if entry.priority_fee_merkle_root_upload_authority != MerkleRootUploadAuthority::Unset {
+            return Err(ValidatorHistoryError::PriorityFeeDistributionAccountAlreadyCopied.into());
+        }
+    }
+
     let mut pdfa_data: &[u8] = &ctx.accounts.distribution_account.try_borrow_data()?;
 
-    let distribution_account = PriorityFeeDistributionAccount::try_deserialize(&mut pdfa_data)?;
+    let distribution_account =
```
coachchucksol, line 84: Could we make this a bit more verbose:

let (commission_bps, priority_fees_transferred, merkle_root_upload_authority) = if PriorityFeeDistributionAccount.dne { (0, 0, DNE_AUTHORITY) } else { 
(distribution_account.validator_commission_bps, distribution_account.total_lamports_transferred, distribution_account.merkle_root_upload_authority)
}

I don't want to rely on default here, I want to expressly zero out all of the data

coachchucksol (changes requested):
Another thought

programs/steward/src/score.rs
```diff
@@ -522,34 +522,56 @@ pub fn calculate_priority_fee_commission(
...
-                max_priority_fee_commission = commission_bps;
-                max_priority_fee_commission_epoch = max_commission_epoch;
-            }
-            Ok(commission_bps)
-        })
-        .collect::<Result<Vec<u16>>>()?;
+        .flat_map(
+            |(relative_epoch, ((tips, total_fees), priority_fee_merkle_root_upload_authority))| {
+                let commission_bps: u16 = calculate_realized_commission_bps(tips, total_fees);
+                // this should not happen, but we should not score the epoch if the upload authority is None
```
ebatsell, line 540: the way the `_range` function works this would actually return None for `MerkleRootUploadAuthority::Unset` since it's representing the same thing -- _range returns None when the value is equal to u8::MAX since that's how we specify unset values for all ValidatorHistory fields

programs/steward/src/score.rs
```diff
@@ -522,34 +522,56 @@ pub fn calculate_priority_fee_commission(
...
+                    return vec![];
+                }
+                if let Some(upload_authority) = priority_fee_merkle_root_upload_authority {
+                    // do not include this epoch in scoring if the upload authority is Unset
+                    if matches!(upload_authority, MerkleRootUploadAuthority::Unset) {
+                        return vec![];
+                    }
+                    // commission is 100% if the PFDA does not exist, validator keeps all fees
+                    if matches!(upload_authority, MerkleRootUploadAuthority::DNE) {
+                        return vec![BASIS_POINTS_MAX];
```
ebatsell, line 551: you could do commission_bps = BASIS_POINTS_MAX here so it goes through the if statement below too

programs/validator-history/src/instructions/copy_priority_fee_distribution.rs
```diff
@@ -57,20 +56,53 @@ pub fn handle_copy_priority_fee_distribution_account(
...
+        }
+    }
+
     let mut pdfa_data: &[u8] = &ctx.accounts.distribution_account.try_borrow_data()?;
 
-    let distribution_account = PriorityFeeDistributionAccount::try_deserialize(&mut pdfa_data)?;
+    let distribution_account = PriorityFeeDistributionAccount::try_deserialize(&mut pdfa_data)
+        .unwrap_or(PriorityFeeDistributionAccount {
+            validator_vote_account: Pubkey::default(),
+            merkle_root_upload_authority: Pubkey::default(),
```
coachchucksol, line 87: nit: Set as DNE_AUTHORITY not Pubkey::default(),

tests/tests/steward/test_scoring.rs
```diff
@@ -709,6 +895,20 @@ mod test_calculate_priority_fee_commission {
                 100,
                 100,
             ],
+            &[
+                MerkleRootUploadAuthority::TipRouter,
+                MerkleRootUploadAuthority::TipRouter,
+                MerkleRootUploadAuthority::TipRouter,
```
coachchucksol, line 901: Any test cases with MerkleRootUploadAuthority::Other?

coachchucksol (changes requested):
A couple more things

PR #142, opened 2025-08-05
steward: emit average pf commission rather than max

gzalz:
**Problem**
We are emitting the max priority fee commission in ScoreDetails, yet we use average priority fee commission within the scoring calculation.

**Solution**
- Do not fail fast when current epoch < go live epoch, perform all calculations, default score to 1.0 and epoch to DEFAULT_EPOCH while returning actual avg_commission
- Return avg_commission rather than max commission when calculating priority fee commission score for a validator
- Update relevant tests and variable names within tests

### DIFF
--- a/programs/steward/idl/steward.json
+++ b/programs/steward/idl/steward.json
@@ -3151,7 +3151,7 @@
           {
             "name": "avg_priority_fee_commission",
             "docs": [
-              "Max realized priority fee commission observed"
+              "Average realized priority fee commission observed"
             ],
             "type": "u16"
           },


programs/steward/src/score.rs
```diff
@@ -98,7 +98,7 @@ pub struct ScoreDetails {
     pub max_historical_commission_epoch: u16,
 
     /// Max realized priority fee commission observed
```
alex-at-jito-sol, line 100: Docstring

programs/steward/src/score.rs
```diff
@@ -585,19 +582,14 @@ pub fn calculate_priority_fee_commission(
     let avg_commission: u16 = u16::try_from(avg_commission).map_err(|_| ArithmeticError)?;
 
     let max_commission = config.max_avg_commission();
-
+    // We would still like to emit avg_commission before the go-live epoch
+    if current_epoch < config.parameters.priority_fee_scoring_start_epoch {
+        return Ok((1.0, avg_commission, EPOCH_DEFAULT));
```
ebatsell, line 587: You can return max_priority_fee_commission_epoch here, somewhat informative for verifying calculation was done right

ebatsell (approved):

PR #145, opened 2025-08-05
feat: use priority fee merkle root upload authority as scoring component

gzalz:
**Problem**

The priority fee merkle root upload authority is not yet a component of the score despite appearing to be in the unstake and scoring details.

**Solution**
- Include priority fee merkle root upload authority as a component of scoring
- Ensure that an Unset authority results in a score of 1.0 to default to not penalizing the validator in this scenario

### DIFF
--- a/tests/tests/steward/test_scoring.rs
+++ b/tests/tests/steward/test_scoring.rs
@@ -591,6 +591,94 @@ mod test_calculate_realized_commission_bps {
     }
 }
 
+mod test_calculate_priority_fee_merkle_root_upload_authority {
+    use super::*;
+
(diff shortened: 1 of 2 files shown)


tests/tests/steward/test_scoring.rs
```diff
@@ -591,6 +591,94 @@ mod test_calculate_realized_commission_bps {
     }
 }
 
+mod test_calculate_priority_fee_merkle_root_upload_authority {
+    use super::*;
+
```
gzalz, line 596: Is it problematic both MEV and PF merkle root authority scoring only consider the latest epoch despite a larger scoring window than one epoch?

PR #148, opened 2025-08-06
CLI: default value should be null

aoikurokawa:
- Some of the fields are added recently, so when i run the command, the default value is displayed. We should print out `null` instead.


coachchucksol (dismissed):
Great change!

PR #149, opened 2025-08-06
fix: Use cluster shred version from solana_net_utils

gzalz:
**Problem**
We are unable to receive data on the gossip network using a shred version of 0 since hitting a jito-solana v2.3 validator.

**Solution**
- Emit a metric for crds_nodes, giving visibility into the amount of nodes we get gossip data on
- Use solana_net_utils to fetch the appropriate cluster shred version to advertise on our gossip node
- Remove cruft related to fully deprecated gossip message types: LegacyVersion, Version, and LegacyContactInfo

```

```

PR #151, opened 2025-08-07
Keeper: Accept multiple gossip entry URLs

aoikurokawa:
- Accept multiple gossip entry URLs

```
# .env file

GOSSIP_ENTRYPOINTS=ENTRY_URL1,ENTRY_URL2,ENTRY_URL3
```

- If we can not get entry (0), try another url to retrieve entry info


aoikurokawa:
Merge after #149

PR #165, opened 2025-08-28
steward: Add directed stake types to steward state

gzalz:
**Problem**
We would like to begin calculating directed stake meta targets off-chain but require on-chain directed staking types to do so.

**Solution**
* Add `DirectedStakePreference` to tracks pairs of delegation % and validator pubkey
* Add `DirectedStakeTicket` to track token holder preferences and token holder type
* Add `DirectedStakeTarget` to represent a directed staking target delegation and track total lamports applied for the direct delegation in question
* Add `DirectedStakeMeta` to store all of the `DirectedStakeTarget` for a given epoch and track on-chain data population progress


programs/steward/src/state/directed_stake.rs
```diff
@@ -0,0 +1,244 @@
+use std::mem::size_of;
+
+use crate::errors::StewardError::{
+    AlreadyPermissioned, DirectedStakeStakerListFull, DirectedStakeValidatorListFull,
+    StakerNotInWhitelist, ValidatorNotInWhitelist,
+};
+use crate::utils::U8Bool;
+use anchor_lang::prelude::*;
+use borsh::BorshSerialize;
+
```
gzalz, line 10: Looking for suggestions on these constants.

programs/steward/src/state/directed_stake.rs
```diff
@@ -0,0 +1,240 @@
...
+        }
+        allocations
+    }
+}
+
+#[derive(BorshSerialize)]
+#[account(zero_copy)]
+pub struct DirectedStakeWhitelist {
+    pub permissioned_stakers: [Pubkey; MAX_PERMISSIONED_DIRECTED_STAKERS],
+    pub permissioned_validators: [Pubkey; MAX_PERMISSIONED_DIRECTED_VALIDATORS],
```
ebatsell, line 128: I like this approach, but wdyt would be better/worse about individual whitelist PDAs? where do you see this account being used?

programs/steward/src/state/directed_stake.rs
ebatsell: We could go higher, maybe 5-10k on stakers, since it's just a singleton account

programs/steward/src/state/directed_stake.rs
```diff
@@ -0,0 +1,240 @@
...
+    pub total_applied_lamports: u128,
+    // Alignment compliant reserve space for future use
+    _padding0: [u8; 64],
+}
+
+#[derive(BorshSerialize)]
+#[account(zero_copy)]
+pub struct DirectedStakePreference {
+    /// Validator vote pubkey
+    pub vote_pubkey: Pubkey,
```
ebatsell, line 67: Consider ValidatorHistory index instead of vote account to save a lot of space here

programs/steward/src/state/directed_stake.rs
ebatsell: I wonder if we should artificially restrict number of preferences to encourage people to have a smaller set of validators, for stickier stake

programs/steward/src/state/directed_stake.rs
```diff
@@ -0,0 +1,240 @@
...
+    pub total_permissioned_validators: u16,
+    // 256 bytes reserved for future use
+    pub _padding0: [u8; 256],
+}
+
+impl DirectedStakeWhitelist {
+    pub const SIZE: usize = 8 + size_of::<Self>();
+    pub const SEED: &'static [u8] = b"whitelist";
+
+    pub fn is_staker_permissioned(&self, staker: &Pubkey) -> bool {
```
ebatsell, line 139: Consider an optional index here (and all ixs that do a scan with pubkey comparisons) for optimized CUs lookup

PR #178, opened 2025-10-15
Upload Validator Age: Relax Constrant

mrmizz:
Do not assert owner of vote account is vote program. 

This vote account could be closed, or even realloced as belonging to a new program. Regardless, it was used to seed a validator history account, and we still want to be able to upload validator age data here. This is safe because the instruction does not read any state from the vote account.

### DIFF
--- a/programs/validator-history/src/instructions/upload_validator_age.rs
+++ b/programs/validator-history/src/instructions/upload_validator_age.rs
@@ -2,7 +2,8 @@ use crate::{
     errors::ValidatorHistoryError,
     state::{Config, ValidatorHistory},
 };
-use anchor_lang::prelude::*;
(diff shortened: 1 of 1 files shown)


programs/validator-history/src/instructions/upload_validator_age.rs
```diff
@@ -13,8 +13,7 @@ pub struct UploadValidatorAge<'info> {
     )]
     pub validator_history_account: AccountLoader<'info, ValidatorHistory>,
 
-    /// CHECK: fine since we are not deserializing account
-    #[account(owner = vote::program::ID.key())]
+    /// CHECK: This account may be closed
```
ebatsell: actually a bit safer to do `vote program || system program`. I think we have this somewhere else...

PR #185, opened 2025-10-17
feat: Directed staking onchain

gzalz:
# JitoSOL Directed Staking

## Overview
Directed staking enables a permissioned set of stakers to delegate the SOL underpinning their JitoSOL balance towards specific permissioned validator preferences. Directed staking delegations will occur at the start of every epoch and the source-of-truth for directed stake delegation target amounts will be a permissioned upload operated by Jito Labs. Certain constraints such as per-epoch un-staking caps and an undirected stake TVL floor are applied to prevent yield drag and an imbalance of the stake pool in a way that dis-proportionately favors directed stake.

### Steward State Machine
- **RebalanceDirected** at 0% epoch progress of every epoch  
- **Idle** when `REBALANCE_DIRECTED_COMPLETED` flag is set  
- **ComputeScores** at 50% epoch progress of scoring epoch  
- **ComputeDelegations** when `COMPUTE_SCORES` flag is set during scoring epoch  
- **Idle** when `COMPUTE_DELEGATIONS` set during scoring epoch  
- **ComputeInstantUnstake** at 90% of epoch progress every epoch  
- **Rebalance** when `COMPUTE_INSTANT_UNSTAKE` flag set every epoch  
- **Idle** when StewardState progress is complete every epoch  
- When a new epoch has been detected the `EpochMaintance` will reset state to `RebalanceDirected` and clear per-epoch flags

### Authorities
- **`directed_stake_whitelist_authority`**  
  The permissioned authority for instructions that add or remove permissioned entities from the whitelist. Practically speaking, this will be a multi-sig address.

- **`directed_stake_meta_upload_authority`**  
  The permissioned authority for instructions that copy target values to the directed stake meta. Practically speaking, this will be a hot wallet running on our stakenet keeper instances.

### Parameters
- **`directed_unstake_cap_bps`**  
  The maximum percent of stake pool TVL allowed to be unstaked from directed staking targets in a single epoch.

- **`undirected_stake_floor_lamports`**  
  The minimum undirected stake TVL in-order to permit directed stake rebalance increases.

### Steward State
- **`directed_unstake_total`**  
  Tracks the amount of directed stake that has been unstaked in the current epoch. This is reset when the state-machine transitions out of the state `RebalanceDirected` to ensure that it starts as zero at the start of each epoch

## Directed Stake Whitelist
The directed stake whitelist is the source-of-truth for permissioned stakers and validators. This list differentiates between user directed stakers and protocol directed stakers for future off-chain meta calculations and business purposes. Instructions to add and remove stakers and validators are permissioned by the `directed_stake_meta_whitelist_authority`.

## Directed Stake Ticket
A directed stake ticket declares a permissioned staker’s delegation preferences. Delegation preferences must be made up of permissioned validators and the total number of allocations per-ticket cannot exceed 10,000 basis points.

## Directed Stake Meta
The directed stake meta is the source-of-truth for validator target delegation amounts and tracking the number of lamports that have been staked with that validator attributable to directed staking. Each stake target within the stake meta will track: the vote pubkey of the validator, the target delegation amount, the staked amount, the epoch the target was last updated, and the epoch the target was last rebalanced (and staked amount last updated).

### Directed Stake Whitelist Lifecycle Instructions
- Initialize  
- Realloc  

### Directed Stake Ticket Lifecycle Instructions
- Initialize  
- Close  

### Directed Stake Meta Lifecycle Instructions
- Initialize  
- Realloc  

### Directed Stake Ticket Update Instruction
- `UpdatedDirectedStakeTicket`

### Directed Stake Meta Copy Instruction
- `CopyDirectedStakeTargets`

### Rebalance Instructions
- `Rebalance`  
- `RebalanceDirected`


programs/steward/src/state/steward_state.rs
```diff
@@ -225,6 +232,22 @@ impl IdlBuild for StewardStateEnum {
                         name: "Rebalance".to_string(),
                         fields: None,
                     },
+                    IdlEnumVariant {
+                        name: "RebalanceDirected".to_string(),
+                        fields: None,
+                    },
+                    IdlEnumVariant {
```
gzalz: Delete these old state enum entries

programs/steward/src/state/steward_state.rs
```diff
@@ -360,15 +393,20 @@ impl StewardState {
...
         let completed_loop = self.has_flag(REBALANCE);
+        let completed_directed_rebalance = self.has_flag(REBALANCE_DIRECTED);
 
         if current_epoch >= self.next_cycle_epoch {
             self.reset_state_for_new_cycle(
                 current_epoch,
                 current_slot,
                 num_epochs_between_scoring,
             )?;
+        } else if !completed_directed_rebalance {
```
gzalz: This is now the init state. Transition from idle not required. We need to determine how we transition from Idle to ComputeScores/Delegations. This should be based on epoch progress.

programs/steward/src/directed_delegation.rs
```diff
@@ -0,0 +1,174 @@
...
+    }
+
+    let vote_pubkey = directed_stake_meta.targets[target_index].vote_pubkey;
+    let target_lamports = directed_stake_meta
+        .get_target_lamports(&vote_pubkey)
+        .ok_or(StewardError::ValidatorIndexOutOfBounds)?;
+
+    let delta_lamports = target_lamports.saturating_sub(current_lamports);
+
+    if delta_lamports == 0 {
```
gzalz: The current lamports check at the end could be moved up here

programs/steward/src/instructions/rebalance_directed.rs
```diff
@@ -0,0 +1,368 @@
...
+                    ctx.accounts.system_program.to_account_info(),
+                    ctx.accounts.stake_program.to_account_info(),
+                ],
+                &[&[
+                    StewardStateAccount::SEED,
+                    &ctx.accounts.config.key().to_bytes(),
+                    &[ctx.bumps.state_account],
+                ]],
+            )?;
+            msg!("decrease_validator_stake_with_reserve successful");
```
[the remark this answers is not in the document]
gzalz: Add even more information. How many llamports? What validator? We can debug better with more information.

tests/tests/steward/test_directed_increase_decrease.rs
```diff
@@ -0,0 +1,482 @@
...
+        1_000_000,
+        5_000_000,
+        1_000,
+        2_000_000,
+    );
+
+    assert!(result.is_ok());
+    match result.unwrap() {
+        RebalanceType::Decrease(components) => {
+            assert!(components.directed_unstake_lamports > 0);
```
gzalz: Exact amounts plz!

programs/steward/src/state/directed_stake.rs
```diff
@@ -0,0 +1,404 @@
...
+            if target.staked_last_updated_epoch != epoch {
+                return false;
+            }
+        }
+        true
+    }
+}
+
+#[derive(BorshSerialize)]
+#[account(zero_copy)]
```
aoikurokawa: i think don't need account macro, reference from ValidatorHistoryEntry
https://github.com/jito-foundation/stakenet/blob/feb71e7e6a4e90ea6e7ed4a5bfe905bb547224cc/programs/validator-history/src/state.rs#L153

programs/steward/src/instructions/close_directed_stake_ticket.rs
```diff
@@ -0,0 +1,48 @@
...
+#[derive(Accounts)]
+pub struct CloseDirectedStakeTicket<'info> {
+    #[account()]
+    pub config: AccountLoader<'info, Config>,
+
+    #[account(
+        mut,
+        close = authority
+    )]
+    pub ticket_account: AccountLoader<'info, DirectedStakeTicket>,
```
mrmizz: I believe we need to add the config pubkey to the derivation of the ticket_account. Here, and in the respective init instruction for ticket accounts.

It seems that there is a bug here, where someone could pass in a "rogue" config account, and close this ticket account.

The flow would be like this. A random keypair goes to the spl-stake-pool program and creates a stake pool. From what I can tell, that is a permissionless instruction. Then they would go to the steward program and create a new steward account and respective config. That config would specify their own `.directed_stake_whitelist_authority`. And then pass that authority in as a signer here.

I've scanned the steward program, and this ticket account is the only PDA that does not use the config pubkey as a seed.

PR #200, opened 2025-11-11
CLI, Keeper: RebalanceDirected

aoikurokawa:
### Problem

In the off-chain get_directed_staking_validators function, iterating through all validators from the stake pool's validator list instead of iterating through the target validators from the directed stake metadata. This caused an edge case failure when a validator was previously in the stake pool validator list but had since been removed.

### Solution

Changed the approach to send instructions for all target validators from the directed stake metadata. When a target validator doesn't exist in the current validator list:

Pass u64::MAX as the validator_list_index
Pass Pubkey::new_unique() as the transient_address

This way, you handle both active validators and those that have been removed from the stake pool validator list.


gzalz (approved):
LGTM. Thanks for tackling

```
> Program logged: "Instruction: RebalanceDirected"> Program logged: "Validator no longer apart of validator list, marking as rebalanced for this epoch."> Program consumed: 19892 of 1399850 compute units
```

PR #232, opened 2025-12-04
Some cli improvements

ebatsell:
Adds commitment to CLIs and improves the "active set" logic to do what we really want: only cranking stake pool vote accounts

### DIFF
--- a/utils/steward-cli/src/commands/command_args.rs
+++ b/utils/steward-cli/src/commands/command_args.rs
@@ -1,7 +1,26 @@
-use clap::{arg, command, Parser, Subcommand};
+use clap::{arg, command, Parser, Subcommand, ValueEnum};
 use jito_steward::{UpdateParametersArgs, UpdatePriorityFeeParametersArgs};
-use solana_sdk::pubkey::Pubkey;
+use solana_sdk::{commitment_config::CommitmentConfig, pubkey::Pubkey};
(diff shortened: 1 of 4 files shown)


utils/steward-cli/src/commands/command_args.rs
```diff
@@ -57,6 +76,10 @@ pub struct Args {
     #[arg(long, global = true, env)]
     pub signer: Option<String>,
 
+    /// Commitment level for RPC queries
+    #[arg(long, global = true, env, default_value = "confirmed")]
+    pub commitment: CommitmentLevel,
+
```
aoikurokawa: can we use this instead?

```rust
use solana_sdk::commitment_config::CommitmentLevel;

impl From<CommitmentLevel> for CommitmentConfig {
    fn from(level: CommitmentLevel) -> Self {
        match level {
            CommitmentLevel::Processed => CommitmentConfig::processed(),
            CommitmentLevel::Confirmed => CommitmentConfig::confirmed(),
            CommitmentLevel::Finalized => CommitmentConfig::finalized(),
        }
    }
}

#[arg(long, global = true, env, default_value = "confirmed", value_parser = clap::value_parser!(CommitmentLevel))]
pub commitment: CommitmentLevel,
```

aoikurokawa (approved):

PR #253, opened 2025-12-30
vote-state: Handle field deserialization for VoteStateV4

gzalz:
**Problem**
SIMD-0185 introduces a new VoteState variant: VoteStateV4. Our VoteStateVersions logic does not handle the deserialization of fields on the V4 variant.

**Solution**
Our VoteStateVersions gracefully handles multiple versions of the VoteState account. The logic used is determined by the enum discriminator which specifies which version of the struct is in use.

- Add support for enum value `[3, 0, 0, 0]`
- Add support for the new enum variant in `deserialize_commission`
- Add support for the new enum variant in `deserialize_epoch_credits`
- Add support for the new enum variant in `deserialize_node_pubkey`

### DIFF
--- a/utils/vote-state/src/lib.rs
+++ b/utils/vote-state/src/lib.rs
@@ -365,8 +365,6 @@ impl VoteStateVersions {
             3 => {
                 let bls_key_option_variant_idx: usize = Self::ENUM_LEN_BYTES
                     + (4 * Self::PUBKEY_BYTES)
-                    + Self::ENUM_LEN_BYTES
-                    + (4 * Self::PUBKEY_BYTES)
                     + Self::INFLATION_REWARDS_COMMISSION_BPS_BYTES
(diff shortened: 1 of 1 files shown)


utils/vote-state/src/lib.rs
```diff
@@ -471,20 +537,180 @@ mod tests {
...
+            false,
+            &mut lamports,
+            ser_current.as_mut_slice(),
+            &owner,
+            false,
+            0,
+        );
+        let epoch_credits_result =
+            VoteStateVersions::deserialize_epoch_credits(&account_current).unwrap();
+        assert!(epoch_credits_result == test_epoch_credits);
```
aoikurokawa: nit: I think `assert_eq!` is better, when a test fails, assert_eq! gives us a helpful diff showing both values which makes debugging much faster.

```suggestion
        assert_eq!(epoch_credits_result, test_epoch_credits);
```

PR #272, opened 2026-03-10
Validator History Program: Copy is bam connected

aoikurokawa:


programs/validator-history/src/state.rs
```diff
@@ -189,7 +189,12 @@ pub struct ValidatorHistoryEntry {
     pub block_data_updated_at_slot: u64,
     /// The enum mapping of the Validator's Tip Distribution Account's merkle root upload authority
     pub priority_fee_merkle_root_upload_authority: MerkleRootUploadAuthority,
-    pub padding1: [u8; 47],
+
+    /// Whether this validator is running the Jito BAM client.
+    /// 1 = running BAM client, 0 = not running BAM client, u8::MAX = unset/unknown.
+    pub is_jito_bam_client: u8,
```
ebatsell: I think a better name for this field and instruction would be `bam_connected` or `is_bam_connected`

aoikurokawa: Thanks! Changed name to `is_bam_connected`.
[ba4410a](https://github.com/jito-foundation/stakenet/pull/272/commits/ba4410a00ed272483f44c7da4bdd5e66a48341ab)

PR #277, opened 2026-03-14
Steward: `running_bam` score

aoikurokawa:
- Transition to 100% BAM stake
- Related to this [JIP](https://forum.jito.network/t/jip-28-accelerate-bam-adoption/904)


programs/steward/src/score/running_bam.rs
```diff
@@ -0,0 +1,66 @@
...
+        .iter()
+        .filter(|entry| matches!(entry, Some(1)))
+        .count();
+
+    if bam_connected_count >= minimum_epochs as usize {
+        1
+    } else {
+        0
+    }
+}
```
ebatsell, line 25: Can we update the logic here so if BAM status was not uploaded for an epoch, we do not consider that epoch?

something like this:
1) count the `None`s in `is_bam_connected_window`, set to num_missed_epochs
2) update comparison to `bam_connected_count >= minimum_epochs - num_missed_epochs`

aoikurokawa: Thanks! Addressed [e1cb630](https://github.com/jito-foundation/stakenet/pull/277/commits/e1cb630f631e4dfd540ccf5623987ab9a3f0940e)

ebatsell (approved):
Looks good. As we observe the first couple of epochs, we might want to tweak it more so we're not just treating missed epochs as successes, but I think this protects us against bad scenarios while still being a good way to track status

aoikurokawa:
Thank you for reviewing!

PR #294, opened 2026-05-18
Keeper: Delegate directed stake to coinbase validator

aoikurokawa:
Coinbase Validator

```txt
Identity address: DidkDYa64DEMx97P4iahDwsWtgUTnLiEhHSMUbQGihi5
Vote address: DyjoG2US2XiVvL1MVzL2wjS5jCTmvo6aNwvqJfaPNx9u
```


sdk/src/utils/instructions.rs
```diff
@@ -204,3 +205,90 @@ pub async fn compute_directed_stake_meta(
...
+
+    if let Some(cb_balance) = coinbase_balance.coinbase_balance {
+        let mut cb_balance_lamports = cb_balance.cb_balance_lamports;
+        if let Some(validator_list_index) = validator_list_account
+            .validators
+            .iter()
+            .position(|v| v.vote_account_address.eq(coinbase_vote_pubkey))
+        {
+            if let Some(meta_target_lamports) = targets.get(coinbase_vote_pubkey) {
+                cb_balance_lamports = cb_balance_lamports.saturating_add(*meta_target_lamports);
```
ebatsell: why are we adding the existing target lamports to the coinbase balance response from kobe?

ebatsell: should it not just be setting the new target via what we get back from the API?

aoikurokawa: First of all, the keeper sends normal directed delegation (ticketing system) txs here.
https://github.com/jito-foundation/stakenet/blob/52ab11dba7f77da70f30a1728d50713587fdc169/keepers/stakenet-keeper/src/entries/crank_copy_directed_stake_targets.rs#L27-L44

after that, it sends cb delegation, so if there is coinbase public key in the first batch txs, we should set target normal delegation + CB delegation.

aoikurokawa: Will comment out normal tx (ticket system) parts now since we are not using right now, and overwrite target number that is coming from kobe api

ebatsell (approved):

PR #299, opened 2026-05-27
Keeper: Unstake from validator with score 0 after epoch 980

aoikurokawa:


keepers/stakenet-keeper/src/state/keeper_state.rs
```diff
@@ -344,11 +348,19 @@ impl KeeperState {
...
         if epoch_progress <= 0.5 {
             log::debug!(
                 "Epoch progress {:.2}% - too early to copy targets",
-                epoch_progress * 100.0
+                epoch_progress * 100.0,
+            );
+            return Ok(false);
+        }
+
+        if is_scoring_epoch && !steward_inner.has_flag(jito_steward::COMPUTE_DELEGATIONS) {
```
ebatsell: next_cycle_epoch gets set pretty early in the epoch so I'm not sure we would get to this point in the epoch where `is_scoring_epoch` == true

sdk/src/utils/instructions.rs
ebatsell: do we need to filter zero-score instructions here too?

aoikurokawa: Good catch, we should filter as well 👍

aoikurokawa: Addressed [a93ea12](https://github.com/jito-foundation/stakenet/pull/299/commits/a93ea128fef46564cbab63dc1f66b0a9f825979d)

keepers/stakenet-keeper/src/state/keeper_state.rs
aoikurokawa: Right! I changed the evaluation from `is_scoring_epoch` to state tag 🫡 
Addressed! [f2763fb](https://github.com/jito-foundation/stakenet/pull/299/commits/f2763fb28c10fb249ac7c8be70921bb5785d6a16)

ebatsell (approved):

PR #305, opened 2026-06-03
Keeper: Skip copying directed stake target

aoikurokawa:
- Skip copying copying directed stake target

### DIFF
--- a/.github/workflows/build-container-images.yaml
+++ b/.github/workflows/build-container-images.yaml
@@ -4,9 +4,9 @@ on:
   workflow_dispatch:
     inputs:
       service_config:
-        description: 'Service configuration'
+        description: "Service configuration"
         required: true
-        default: 'stakenet-keeper'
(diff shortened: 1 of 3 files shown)


.github/workflows/build-container-images.yaml
```diff
@@ -18,38 +18,37 @@ env:
...
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Extract metadata
+        id: meta
+        uses: docker/metadata-action@v5
+        with:
+          images: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
+          tags: |
+            type=raw,value=${{ github.ref_name }}-${{ github.event.inputs.service_config }}-{{sha}}
+
```
aoikurokawa, line 43: remove latest-tag here since we are not using

gzalz (approved):

PR #322, opened 2026-08-05
CLI: Add `view-stake-eta` to estimate when a validator receives stake

john-jito:
## Problem

Validators can already see **whether** they are eligible on the [steward page](https://www.jito.network/stakenet/steward/). What they cannot see is *when to expect stake*, or — if they are in the set but below target — *why nothing is happening*.

The premise that this can't be answered programmatically turns out to be too strong. The wait is deterministic in almost every input; only the arrival date depends on exogenous deposit flow.

## What this adds

`steward-cli view-stake-eta` answers it from public on-chain state, with no program or governance change. Four accounts in one `getMultipleAccounts`.

Increases are funded from the reserve in **descending score order**, so a validator's wait is driven by the cumulative shortfall of everyone ranked above it, and by how much rotation budget is left in the cycle (`scoring_unstake_cap_bps`, reset every `num_epochs_between_scoring` epochs). The command reports rank, SOL ahead in the queue, remaining budget and its reset epoch, and a confidence-banded ETA — led by a `limiting_factor` that names the actual blocker (`queue_position`, `scoring_churn_budget_exhausted`, `reserve_empty`, `next_rebalance`, …) rather than just an epoch count.

Two extra modes:

- `--top N` — the validators furthest over target, and the largest holders of directed stake. Directed stake is excluded from algorithmic targets, so this makes clear how much of the pool's apparent excess is directed and therefore *not* algorithmically reducible.
- `--schedule-epochs N` — projects movement forward, replaying the program's ordering rules: decreases lowest-raw-score first against the cycle budgets, increases highest-score first out of the reserve, one epoch of cooldown between them. `--min-change-sol` drops the long tail of trivial adjustments.

## The subtlety worth reviewing

Targets divide the pool **net of directed stake**, matching `instructions/rebalance.rs:259`. Dividing the gross pool instead overstates every target by `directed_total / N` — currently ~1,218 SOL per validator — which aggregates into a phantom pool-wide shortfall equal to the entire directed balance, and looks convincingly like a structural problem rather than an arithmetic slip. An earlier draft of this work made exactly that error.

Because the failure mode is a *silently plausible wrong number* rather than a crash, an unreadable `DirectedStakeMeta` is now fatal rather than defaulting to zero directed stake. `--assume-no-directed-stake` overrides it for deployments that genuinely have none.

## Docs

`utils/steward-cli/STAKE_ETA.md` documents the model — why arrival is a queue rather than a schedule, the exact inputs, the output schema, the reconciliation invariant, and the known limitations. Linked from the CLI README with a note to read it before changing the estimator.

## Testing

28 unit tests. Four drive the real `StewardStateV2::rebalance` and assert the program's own increase/decrease amounts match the model:

- the program funds an empty validator to exactly the model's target, and *not* the gross-pool figure
- an over-target validator is unstaked by exactly the model's excess, in the `scoring` category
- a validator's own directed stake does not count toward its target (opposite sign to what it would be if it did, so it cannot pass by accident)
- the projection's first epoch reproduces the program instruction-for-instruction

Coverage alone proves little, so each of these faults was injected deliberately and confirmed to fail tests:

| Injected fault | Tests killed |
| :- | -: |
| Drop the directed-stake subtraction | 5 |
| Derive rank from index order, not score order | 2 |
| Remove the one-epoch cooldown | 2 |
| Never reset the cycle budget at the boundary | 1 |
| Ignore the minimum-delegation floor when queueing | 1 |

## Known limitations

Documented in STAKE_ETA.md §8 rather than left implicit:

- Deposit flow is exogenous and is the dominant uncertainty; the low bound assumes none, and the optimistic bound is operator-supplied via `--deposit-rate-sol` rather than derived from keeper metrics yet.
- Multi-epoch projection beyond the first epoch is property-tested (cycle reset, cooldown, convergence) rather than verified against the program, which would need a stake-pool simulator applying transient stake between epochs.
- Within a cycle the unstake budget is not rate-limited per epoch, so the projection spends it as soon as a cycle opens. Per-cycle totals are reliable; the exact epoch within a cycle is indicative, since real timing depends on when cranks run.
- Any estimate crossing a cycle boundary assumes the set and N are unchanged; both are re-derived at each boundary.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/utils/steward-cli/README.md
+++ b/utils/steward-cli/README.md
@@ -105,7 +105,12 @@ cargo run -p steward-cli -- \
     --min-change-sol 5000
 ```
 
-The projection prints per-epoch movements, a per-validator rollup, and per-epoch totals.
+The projection prints per-epoch movements, a per-validator rollup, and per-epoch totals. The
+rollup separates **undirected** from **directed** balance, and that distinction matters when
+reading it: targets are measured against undirected stake only, so a validator holding directed
+stake shows a near-zero undirected figure and a full-target shortfall while actually holding
+several times the target. Those rows are flagged inline with the real total, so they are not
+mistaken for the pool's most underfunded validators.
 Because the unstake budget is cycle-scoped rather than rate-limited per epoch, the projection
 spends it as soon as a cycle opens; treat the per-cycle totals as reliable and the exact epoch
 within a cycle as indicative, since real timing depends on when cranks run.
--- a/utils/steward-cli/README.md
+++ b/utils/steward-cli/README.md
@@ -45,6 +45,14 @@ cargo run -p steward-cli -- \
 Estimates how many epochs until a validator receives stake, given current pool conditions.
 Answers "when should I expect stake?" for a validator that is already eligible.
 
+> **This does not guarantee stake.** The command reports what is firm (eligibility, target, rank,
+> queue position, remaining budget — all read from chain) alongside what is a projection of
+> current conditions. Deposit and withdrawal flow, validator performance, other operators'
+> commission changes, and the re-scoring at every cycle boundary are not predictable and are not
+> modelled. A figure 10 or 20 epochs out can differ substantially from what happens. The output
+> carries this caveat in both the human banner and the JSON `disclaimer` field — **if you share
+> this with validators, the caveat has to travel with it**, or a projection gets read as a promise.
+
 **[STAKE_ETA.md](./STAKE_ETA.md) documents the model this implements** — why arrival is a queue
 rather than a schedule, the exact inputs, the output schema, and the known limitations. Read it
 before changing the estimator.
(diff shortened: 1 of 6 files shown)


john-jito:
Closed in favour of the documentation approach in #324, which publishes the model as agent-readable reference material rather than as a first-party command.

Branch `jh/cli-view-stake-eta` is retained rather than deleted, so this diff stays available.

Technical notes worth keeping:

- Algorithmic targets divide the pool **net of directed stake** (`instructions/rebalance.rs:259`). Dividing gross TVL overstates every target by `directed_total / N`.
- Reconciliation invariant: `total_shortfall - total_excess == reserve + transient`.
- Funding order is descending `score`; unstaking order is ascending `raw_score`.
- 28 tests, four of which drive `StewardStateV2::rebalance` directly and assert the program's own increase/decrease amounts match the model.

All of the above is carried into `agent-guides/` in #324.

PR #328, opened 2026-08-26
fix: enforce owner check on priority fee distribution account (#293)

### LINKED ISSUE #293: priority fee distribution account owner is unchecked
**describe the bug**
`copy_priority_fee_distribution` checks the expected pda seeds for `distribution_account`, but does not check that the account is owned by the configured priority fee distribution program.

code:
https://github.com/jito-foundation/stakenet/blob/0a708371f5859dabe5df0902dfc565add7282007/programs/validator-history/src/instructions/copy_priority_fee_distribution.rs#L34-L107

if a caller passes the correct pda address before the real keeper copy, deserialization can fail and the instruction records the account as `dne`. that poisoned value can persist in validator history.

**impact**
steward can later treat the validator as having 100% realized priority fee commission and/or a bad priority fee upload authority, reducing score or contributing to unstake decisions once priority fee scoring is active.

**note:** on the current jitosol steward config this appears effectively disabled today because `priority_fee_scoring_start_epoch` is in the future and the configured max commission plus error margin is above the 100% value recorded for `dne`. the current risk is persistent validator-history poisoning and future delegation impact if priority fee scoring is enabled while poisoned epochs remain in lookback.

related scoring paths:
https://github.com/jito-foundation/stakenet/blob/0a708371f5859dabe5df0902dfc565add7282007/programs/steward/src/score.rs#L701-L720

**expected behavior**
`distribution_account` should be required to be owned by `config.priority_fee_distribution_program`.

**suggested fix**

```rust
#[account(owner = config.priority_fee_distribution_program.key())]
pub distribution_account: uncheckedaccount<'info>,
```

ayushsingh82:
## Summary

Fixes #293.

`CopyPriorityFeeDistribution` validates `distribution_account`'s PDA seeds but never checks its **owner**, unlike the equivalent `tip_distribution_account` check in `CopyTipDistributionAccount` (which already has `owner = config.tip_distribution_program.key()`).

Because the check was missing, anyone could call this instruction with the correct PDA address *before* the real `PriorityFeeDistributionAccount` was created for that validator/epoch. Deserialization then fails, the handler falls through to `unwrap_or(default)`, and a `DNE` entry gets written permanently — `PriorityFeeDistributionAccountAlreadyCopied` then blocks any later correction once the real account does exist, poisoning validator history for that epoch.

## Fix

Add `owner = config.priority_fee_distribution_program.key()` to `distribution_account`, matching the pattern already used by the tip-distribution sibling instruction. The instruction now fails fast (`ConstraintOwner`) instead of silently recording a poisoned default when the distribution account doesn't exist yet.

## Test plan

- Updated `test_priority_fee_distribution_account_does_not_exist` → `test_priority_fee_distribution_account_does_not_exist_fails`, asserting `ConstraintOwner` instead of a successful DNE write (mirrors `test_mev_commission_fail` for the tip-distribution sibling).
- Ran the full `test_copy_priority_fee_distribution` suite locally (`cargo build-sbf` + `cargo test`): all 6 tests pass, including the pre-existing success-path and double-copy tests, confirming legitimate flows are unaffected.

## One thing worth flagging

The keeper (`priority_fee_commission.rs::update_priority_fee_commission`) sends this instruction for every validator/epoch where the local entry is `Unset`, without pre-checking the distribution account exists on-chain. After this fix, a validator whose distribution account is never created will get a failing (retried) instruction each crank run instead of a one-time `DNE` write. This mirrors the tip-distribution keeper's existing "always attempt, let failures retry" design, so it's consistent with how the codebase already handles the sibling case — but flagging it in case the DNE-recording behavior was intentionally relied on for `steward` scoring somewhere.

### DIFF
--- a/programs/validator-history/src/instructions/copy_priority_fee_distribution.rs
+++ b/programs/validator-history/src/instructions/copy_priority_fee_distribution.rs
@@ -41,6 +41,7 @@ pub struct CopyPriorityFeeDistribution<'info> {
         ],
         bump,
         seeds::program = config.priority_fee_distribution_program.key(),
+        owner = config.priority_fee_distribution_program.key()
     )]
     pub distribution_account: UncheckedAccount<'info>,
 
--- a/tests/tests/validator_history/test_copy_priority_fee_distribution.rs
+++ b/tests/tests/validator_history/test_copy_priority_fee_distribution.rs
@@ -14,7 +14,12 @@ use validator_history::{Config, MerkleRootUploadAuthority, ValidatorHistory};
 const TIP_ROUTER_AUTHORITY: Pubkey = pubkey!("8F4jGUmxF36vQ6yabnsxX6AQVXdKBhs8kGSUuRKSg8Xt");
 
 #[tokio::test]
-async fn test_priority_fee_distribution_account_does_not_exist() {
+async fn test_priority_fee_distribution_account_does_not_exist_fails() {
+    // Prior to the `owner` constraint on `distribution_account`, calling this instruction
+    // before the real distribution account was created would silently succeed and record a
+    // permanent `DNE` entry (see `PriorityFeeDistributionAccountAlreadyCopied`), poisoning
+    // validator history for that epoch. It must now fail instead, matching the equivalent
+    // uninitialized-account case for `copy_tip_distribution_account` (test_mev_commission_fail).
     let fixture = TestFixture::new().await;
     let ctx = &fixture.ctx;
     fixture.initialize_config().await;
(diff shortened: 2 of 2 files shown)


ayushsingh82:
@ebatsell @aoikurokawa could one of you take a look when you have a chance? Tagging since this touches copy_priority_fee_distribution.rs / copy_tip_distribution_account.rs.

ayushsingh82:
Following up on this — mergeable, and it's a straightforward owner-check fix for #293 (priority-fee distribution account). Happy to add tests or adjust scope if that helps get it reviewed.

PR #337, opened 2026-09-23
Program: Handle tower → alpenglow migration marker in vote account epoch credits

aoikurokawa:
## Motivation

Alpenglow changes what StakeNet reads from vote accounts in two ways that break `copy_vote_account` and steward scoring after activation.

1. A migration marker in epoch_credits
Every vote account credited during the tower → alpenglow migration epoch gets a sentinel entry:

```rust
pub const AG_MIGRATION_EPOCH_CREDIT: (Epoch, u64, u64) = (Epoch::MAX, u64::MAX, u64::MAX);
```

[`votor-messages/src/migration.rs#L78-L79`](https://github.com/jito-foundation/jito-solana/blob/219672f7e83f76d623d9bc01f516e4427b094f7a/votor-messages/src/migration.rs#L78-L79)

It is a *separator*, not a value. The migration epoch keeps real entries on both sides of it — one for its tower-era credits, one for its alpenglow-era credits:

```
..., (849, 1100, 1000), (850, 1150, 1100), (MAX, MAX, MAX), (850, 1150+r, 1150)
                        └── tower half ──┘   └ separator ┘   └─ alpenglow half ─┘
```

2. Epoch credits become reward lamports
After the migration, `epoch_credits` accrue lamports ([`vote_reward.rs#L264-L297`](https://github.com/jito-foundation/jito-solana/blob/219672f7e83f76d623d9bc01f516e4427b094f7a/runtime/src/block_component_processor/vote_reward.rs#L264-L297)) - about 52 SOL per epoch for the median mainnet validator, far above `u32::MAX` (~4.29 SOL). That breaks two things:
- Saturating to `u32::MAX` collides with the `ValidatorHistoryEntry` "unset" sentinel, so the validator reads as not voting: flagged for instant unstake and scored 0. On current mainnet numbers that is every validator above ~15k SOL (651 of 678).
- The steward sums the 30-epoch credits window as `u32`, which overflows within two epochs of large values; with release overflow checks, `compute_score` panics.

## Changes

- Define `AG_MIGRATION_EPOCH_CREDIT` in `deserialize_epoch_credits`, and skip it in the keeper and SDK
- Sum the migration epoch's two entries instead of overwriting, and propagate errors instead of `unwrap()` panics in the epoch-credits map.
- Saturate overflowing credits at `MAX_EPOCH_CREDITS`, both per entry and when summing the split epoch, so a stored value never equals the unset sentinel. The unset value itself is unchanged.


programs/validator-history/src/instructions/initialize_validator_history_account.rs
```diff
@@ -29,8 +29,18 @@ pub fn handle_initialize_validator_history_account(
...
+        if previous_epoch != Some(*epoch) {
+            distinct_epochs = distinct_epochs.saturating_add(1);
+            previous_epoch = Some(*epoch);
+        }
+    }
+
+    if distinct_epochs < MIN_VOTE_EPOCHS {
         return Err(ValidatorHistoryError::NotEnoughVotingHistory.into());
     }
+
```
aoikurokawa, line 44: What Alpenglow changes: the migration epoch can appear twice.
could be like

```txt
..., (850, tower half), [marker], (850, alpenglow half)
```

programs/steward/src/score.rs
```diff
@@ -504,7 +504,11 @@ pub fn calculate_epoch_credits(
         return Err(StewardError::ArithmeticError.into());
     }
 
-    let average_vote_credits = epoch_credits_window.iter().filter_map(|&i| i).sum::<u32>() as f64
+    let average_vote_credits = epoch_credits_window
+        .iter()
+        .filter_map(|&i| i)
+        .map(u64::from)
+        .sum::<u64>() as f64
```
aoikurokawa, line 511: because we store u32::MAX - 1 here as default post AG, we should avoid overflow

